PR #44, opened 2023-03-23
Updated project dependencies

Bartvaderkin:
Note this has code changes because of the update where vng-api-common now uses notification-api-common.

Also we choose to only upgrade vng-api-common to 1.9.0 because later version switched from drf-yasg to drf-spectacular and we're not ready for that work (and 1.9.0 supports for Django 3.2)

I tried to split the work over various commits to make it slightly more sane.


stevenbal (approved):
Looks good to me. I tried testing notifications locally, but I think notifications aren't actually properly enabled yet, because `bin/celery_worker.sh` doesn't work? (They may have never been enabled, I'm not sure, it's been too long since I worked on this project :sweat_smile:)

PR #45, opened 2023-03-24
Updated docs and URLs to use new Github location

Bartvaderkin:
I'm not 100% if I got all references that use the repository or cloned directory name (which also changed from `openklant` to `open-klant`)


alextreme (approved):
Appreciated, thanks!

PR #47, opened 2023-04-07
Added identificatie, type, onderwerp, status, antwoord fields to ContactMoment

Bartvaderkin:
For review purposes


src/openklant/components/contactmomenten/openapi.yaml
```diff
@@ -448,35 +483,10 @@ paths:
...
       responses:
         '200':
           description: OK
           headers:
-            ETag:
-              description: De ETag berekend op de response body JSON. Indien twee
-                resources exact dezelfde ETag hebben, dan zijn deze resources identiek
-                aan elkaar. Je kan de ETag gebruiken om caching te implementeren.
-              schema:
-                type: string
```
Bartvaderkin, line 479: @stevenbal I need some advice here:

I regenerated the OAS after updating the dependencies and somehow the ETag support information disappeared. 

I traced this back to updating `vng-api-common` from 1.6.4 to 1.9.0, and starting from 1.7.3 this was changed: https://github.com/VNG-Realisatie/vng-api-common/releases/tag/1.7.3

So I was wondering if Etag support ever was a supported or working feature, or if it just showed up in the OAS because of an issue that was fixed from 1.7.3 (and a few later ones), like I can't find any use of the expected related mixins or decorators on the views and models etc in open-klant.

stevenbal: I vaguely remember that the ETag headers in the OAS were initially regardless of whether the associated resources actually supported caching (indeed the Open Klant resources didn't support this) and that this behavior was indeed changed in 1.7.3.

So I think it's fine to drop these headers :+1:

PR #54, opened 2023-09-04
[#53] Making klantnummer optional and auto-generating klantnummer if not provided

alextreme:


src/openklant/components/klanten/models/klanten.py
```diff
@@ -88,10 +97,19 @@ class Klant(APIMixin, models.Model):
         default=False, help_text=_("Geeft aan of de KLANT wel of niet geverifieerd is.")
     )
 
+    objects = KlantManager()
+
     class Meta:
         verbose_name = "klant"
         verbose_name_plural = "klanten"
-        unique_together = ("bronorganisatie", "klantnummer")
```
annashamray, line 94: I don't quite understand why you remove the DB constraint here
Anyway it looks like the `makemigrations` should be run here

src/openklant/components/klanten/models/klanten.py
```diff
@@ -6,12 +6,21 @@
...
 
 from .constants import GeslachtsAanduiding, KlantType, SoortRechtsvorm
 
 
+class KlantManager(models.Manager):
+    def get_next_klantnummer(self):
+        id_max = Klant.objects.all().aggregate(models.Max("klantnummer"))[
+            "klantnummer__max"
+        ]
+        return int(id_max) + 1 if id_max else 1
```
annashamray, line 21: Since ` klantnummer`  is a char field and it's writable in the API, it's possible that it can be filled with the value which can't be converted into `int` (for example, 'KLANT11')
So if you have such value in DB the code would crush with `ValueError`

src/openklant/components/klanten/api/serializers.py
```diff
@@ -315,6 +316,11 @@ def __init__(self, *args, **kwargs):
                 self.fields[custom_field].help_text
             )
 
+    def validate_klantnummer(self, value):
```
annashamray, line 319: It looks like it duplicates klantnummer generation in the `Klant.save` method

src/openklant/components/klanten/api/tests/test_klant.py
```diff
@@ -348,6 +348,93 @@ def test_create_klant_website_url_optional(self):
         self.assertEqual(klant.bronorganisatie, "950428139")
         self.assertEqual(klant.website_url, "")
 
+    def test_create_klant_website_url_optional_klantnummer(self):
```
annashamray, line 351: Nice tests!

annashamray (approved):
I approve if this issue is urgent and you don't want to wait for my second reviewing.

annashamray:
I also want to comment that this code is prone to race conditions - what if two Klant objects are being created in the same time? 
We had it in Open Zaak and it was very painful, and the solution was also painful. 
If you can avoid id generation on Django level - do it!

src/openklant/components/klanten/models/klanten.py
alextreme: This is because unique_together in DRF makes both fields implicitly required, also in the serializer (so you can't create a klant without having both fields specified):

https://www.django-rest-framework.org/api-guide/validators/#uniquetogethervalidator

> Note: The UniqueTogetherValidator class always imposes an implicit constraint that all the fields it applies to are always treated as required. Fields with default values are an exception to this as they always supply a value even when omitted from user input.

I noticed that this validation was done very early when performing a create, sidestepping the fact that I made klantnummer optional in the serializer. Alternative ideas are welcome, as I would preferably keep the unique_together on a DB level.

src/openklant/components/klanten/api/serializers.py
alextreme: Nice catch, I couldn't get it to work on the serializer so moved this to the save(). Tests should pass without this, so I've removed it.

src/openklant/components/klanten/models/klanten.py
annashamray: In Open Zaak we made a custom validator which replaces the default one. As far as I understand the default validators is used only if you don't have `Meta.validators` in your serializer. But it's more work of course :)

alextreme: migration added

PR #99, opened 2023-10-20
:arrow_up: upgraded all libraries to latest version

bart-maykin:
Updated python/django libaries
Updated Django to 4.1 and set in the install that the python version is 3.10 as latest version requirements of [drf](https://www.django-rest-framework.org/#requirements)


joeribekker (changes requested):
See https://code.djangoproject.com/ticket/33353#comment:13
to solve the build issue.

joeribekker (approved):

Issue #115, opened 2023-12-06
As developer, I want to be sure that Trivy scans the locally built image


joeribekker:
**Tasks**

* [x] Add Trivy, see: https://github.com/open-formulieren/open-forms/pull/3614/files

PR #134, opened 2024-01-11
 🎨 [#115] fixed the trivy image scan 

bart-maykin:


joeribekker (changes requested):
I really want to use the tar-file option to make sure we use the image of this PR and not accidentally download "latest" or somewhat.

bart-maykin:
Proof of passing actions before re-including the "test" job https://github.com/maykinmedia/open-klant/actions/runs/7548469422

joeribekker (approved):

PR #160, opened 2024-01-30
:heavy_plus_sign: Replace dependencies with open-api-framework deps

stevenbal:
taiga issue: https://taiga.maykinmedia.nl/project/maykin-intranet/issue/739


src/openklant/conf/base.py
```diff
@@ -415,21 +415,15 @@
...
 # proxy included.
-AXES_PROXY_COUNT = NUM_PROXIES - 1 if NUM_PROXIES else None
-AXES_ONLY_USER_FAILURES = (
-    False  # Default: False (you might want to block on username rather than IP)
-)
-AXES_LOCK_OUT_BY_COMBINATION_USER_AND_IP = (
-    False  # Default: False (you might want to block on username and IP)
-)
+AXES_IPWARE_PROXY_COUNT = NUM_PROXIES - 1 if NUM_PROXIES else None
+AXES_LOCKOUT_PARAMETERS = ["ip_address"]
```
stevenbal, line 426: @alextreme I noticed that these settings are not the same across different components:

- Open Zaak and Open Klant use `["ip_address"]` (or the equivalent for version 5.x.x)
- Open Forms and Objecttypes use `[["ip_address", "user_agent", "username"]]`

alextreme: @stevenbal default-project uses `AXES_LOCKOUT_PARAMETERS = ["username", "email", "ip"]`, but this may have to do with newer versions of django-axes. As long as the number of attempts is sufficient and the cooldown time isn't too long it shouldn't matter too much, doing blocking at least only on ip_address would have my preference than on the combination with UA + username

stevenbal:
closing in favor of https://github.com/maykinmedia/open-klant/pull/176

PR #185, opened 2024-04-24
:sparkles: [#183] added afgehandeld_op field for internetaken

### LINKED ISSUE #183: Title here
### Thema / Theme

Klantinteracties API

 Als KCM
wil ik kunnen zien wanneer een interne taak is afgehandeld,
zodat ik de klant de juiste informatie kan geven

Het komt helaas nog vaak voor dat klanten nog een keer bellen naar aanleiding van een interne taak (Contactverzoek), die namens hen is uitgezet. Het is op dat moment van belang om te kunnen zien of een taak is afgehandeld, en zoja wanneer, om de klant zo goed mogelijk verder te helpen. 

https://github.com/VNG-Realisatie/klantinteracties/issues/285

bart-maykin:
Fixes #183

- Added `afgehandeld_op` field for `internetaken`
- Automaticly set `afgehandeld_op` when status is set on `verwerkt`
- Automaticly clear `afgehandeld_op` when status is set back on `te_verwerken` if `afgehandeld_op` already had a date


src/openklant/components/klantinteracties/admin/internetaken.py
```diff
@@ -17,9 +17,9 @@ class InterneTaakAdmin(admin.ModelAdmin):
         "nummer",
         "status",
         "toegewezen_op",
+        "afgehandeld_op",
     )
     list_filter = (
         "actor",
         "status",
     )
-    readonly_fields = ("toegewezen_op",)
```
stevenbal, line 25: Was this intentional? It no longer shows up in the detail admin page now

src/openklant/components/klantinteracties/api/serializers/internetaken.py
```diff
@@ -68,20 +71,43 @@ class Meta:
             },
         }
 
+    def validate(self, attrs):
+        status = attrs.get("status", None)
+        if not status and self.instance:
+            status = self.instance.status
+
+        if attrs.get("afgehandeld_op") and status == Taakstatus.te_verwerken.value:
```
stevenbal, line 79: Minor, but I don't think you have to specifically compare to `.value`. Additionally, I'm not sure if new `Taakstatus` values will be added in the future, but to be sure I would probably check if the status is not `verwerkt` instead, to account for that if it does happen.
```suggestion
        if attrs.get("afgehandeld_op") and status != Taakstatus.verwerkt:
```

src/openklant/components/klantinteracties/api/serializers/internetaken.py
```diff
@@ -68,20 +71,43 @@ class Meta:
...
+            raise serializers.ValidationError(
+                {
+                    "afgehandeld_op": _(
+                        "De Internetaak kan geen afgehandeld op datum bevatten "
+                        "als de status nog in '{te_verwerken}' staat."
+                    ).format(te_verwerken=Taakstatus.te_verwerken.value)
```
stevenbal, line 85:
```suggestion
                        "als de status niet op '{verwerkt}' staat."
                    ).format(verwerkt=Taakstatus.verwerkt)
```

src/openklant/components/klantinteracties/api/serializers/internetaken.py
```diff
@@ -68,20 +71,43 @@ class Meta:
...
         klantcontact_uuid = str(validated_data.pop("klantcontact").get("uuid"))
+        afgehandeld_op = validated_data.pop("afgehandeld_op", timezone.now())
 
         validated_data["actor"] = Actor.objects.get(uuid=actor_uuid)
         validated_data["klantcontact"] = Klantcontact.objects.get(
             uuid=klantcontact_uuid
         )
 
+        if validated_data.get("status") == Taakstatus.verwerkt:
+            validated_data["afgehandeld_op"] = afgehandeld_op
```
stevenbal, line 103: Is it necessary to implement this in the serializer `.create` and `.update`? I think it would be easier to implement it in the `InterneTaak` model `.save`.

I think since `afgehandeldOp` is a field for which the value is always `now()` or `None`, it should probably be a read-only field on the serializer, that way you can remove the `.validate` and the changes in `.create` and `.update` and leave the logic in `InterneTaak.save` as is

stevenbal (changes requested):

src/openklant/components/klantinteracties/api/serializers/internetaken.py
bart-maykin: Voor `afgehandeldOp` we want the user to either provide a value or user the date when the request was made. Hence why we shouldn't set the field as read_only. We can definitely set the value if none provided in the model `.save` method

src/openklant/components/klantinteracties/api/serializers/internetaken.py
```diff
@@ -68,6 +70,23 @@ class Meta:
             },
         }
 
+    def validate(self, attrs):
+        status = attrs.get("status", None)
+        if not status and self.instance:
```
Coperh, line 75: Minor thing that is more _"proper"_ 
(I think you are checking if instance is not None) As far as I know, they can only be false if None.

```suggestion
        if status is None and self.instance is not None:
```

Coperh (approved):
Some minor suggestions to but looks good.

src/openklant/components/klantinteracties/api/serializers/internetaken.py
stevenbal: @bart-maykin can you update the error message text as well so it's the same as the conditional?

stevenbal (approved):
Looks good, the final comment about the error message is not super critical, but I think it would be good to keep it in line with the conditional

PR #190, opened 2024-05-17
🔧 [open-zaak/open-zaak#1629] Refactor settings module

stevenbal:
Issue: https://github.com/open-zaak/open-zaak/issues/1629

**Changes**

* Use base settings include from Open API Framework


src/openklant/conf/base.py
```diff
@@ -1,448 +1,40 @@
...
-        "BACKEND": "django_redis.cache.RedisCache",
-        "LOCATION": f"redis://{config('CACHE_AXES', 'localhost:6379/0')}",
-        "OPTIONS": {
-            "CLIENT_CLASS": "django_redis.client.DefaultClient",
-            "IGNORE_EXCEPTIONS": True,
-        },
-    },
-    "oidc": {
-        "BACKEND": "django_redis.cache.RedisCache",
-        "LOCATION": f"redis://{config('CACHE_OIDC', 'localhost:6379/0')}",
```
Coperh, line 92: Another place where `CACHE_OIDC` is mentioned but is different in open-api-framework

src/openklant/conf/base.py
```diff
@@ -1,448 +1,40 @@
...
-    "django.contrib.sites",
-    "django.contrib.messages",
-    "django.contrib.staticfiles",
-    # Admin auth
-    # "django_otp",
-    # "django_otp.plugins.otp_static",
-    # "django_otp.plugins.otp_totp",
-    # "two_factor",
-    # Optional applications.
-    "ordered_model",
```
Coperh, line 121: Again `ordered_model` is not in open-api-framework

Although not used?

src/openklant/conf/base.py
```diff
@@ -1,448 +1,40 @@
...
-    "rest_framework",
-    "django_markup",
-    "solo",
-    "django_jsonform",
-    "log_outgoing_requests",
+# FIXME implement 2fa
+INSTALLED_APPS.pop(INSTALLED_APPS.index("maykin_2fa"))
+INSTALLED_APPS = INSTALLED_APPS + [
     # Project applications.
-    "openklant",
```
Coperh, line 141: Removed and missing from open-api-framework

Coperh: I keep thinking this is OpenZaak not OpenKlant.
This should not have been there in the first place, I dont think.

stevenbal: For some reason adding this breaks things when running tests:

```
django.core.exceptions.ImproperlyConfigured: '/home/steven/Documents/open-klant/src/openklant/fixtures' is a default fixture directory for the 'openklant' app and cannot be listed in settings.FIXTURE_DIRS.
```

annashamray (approved):
Looks good! I'll add also Bart as main Open Klant developer to double check

stevenbal:
@bart-maykin could you do a double check so we can get this merged? :)

Issue #188, opened 2024-04-26
Open Klant has settings for notifications, but notifications are currently unused?


stevenbal:
There are some references to notifications (probably because they were enabled pre v2):
https://github.com/maykinmedia/open-klant/blob/c391829de1a76740e43c69e0eff64dc19bcfa2ed/src/openklant/urls.py#L65
https://github.com/maykinmedia/open-klant/blob/c391829de1a76740e43c69e0eff64dc19bcfa2ed/src/openklant/conf/dev.py#L112
https://github.com/maykinmedia/open-klant/blob/c391829de1a76740e43c69e0eff64dc19bcfa2ed/src/openklant/conf/base.py#L380

I'm not sure if v2 should/will support notifications, but I would prefer if these references are removed if they won't be supported

PR #193, opened 2024-06-07
[#188] remove unused notification settings

Coperh:
Fixes: #188

Other settings removed in #190


stevenbal (approved):

stevenbal (changes requested):
I forgot about this, the pages for both the APIs still show links to the notification docs (which result in a 404), those should be deleted too

image

Coperh:
Remove, but is is displayed on top of each other if that matters. Its technically not docs
Screenshot from 2024-06-07 15-55-50

stevenbal:
@Coperh seems fine to me to display them like that. I will create a separate issue to see what we need to do with the `Runtime behaviour` link though, because it links to documentation by VNG for v1 of the APIs (and also mentions that this is outdated), but I'm not sure if there's any VNG documentation for version 2 of the APIs

stevenbal (approved):

PR #194, opened 2024-06-07
Feature/189 install maykin 2fa

### LINKED ISSUE #189: Replace two_factor with maykin_2fa
I thought that this already happened during the Django 4.2 upgrade, but it seems we're still using our django-two-factor-auth fork (this probably also applies to the other components)

https://maykin-2fa.readthedocs.io/en/stable/

Coperh:
Fixes #189

### DIFF
--- a/src/openklant/accounts/tests/test_admin.py
+++ b/src/openklant/accounts/tests/test_admin.py
@@ -1,0 +1,45 @@
+from django.test import override_settings, tag
+from django.urls import reverse, reverse_lazy
+from django.utils.translation import gettext
+
+from django_webtest import WebTest
+
+from openklant.accounts.tests.factories import RecoveryTokenFactory, SuperUserFactory
+
+LOGIN_URL = reverse_lazy("admin:login")
+
+
+@override_settings(
+    USE_OIDC_FOR_ADMIN_LOGIN=False,
+    MAYKIN_2FA_ALLOW_MFA_BYPASS_BACKENDS=[],  # enforce MFA
+)
+class RecoveryTokenTests(WebTest):
+    @tag("gh-4072")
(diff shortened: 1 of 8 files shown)


Coperh:
Requires maykinmedia/open-api-framework#27 to be merged and released

Coperh:
Requires https://github.com/maykinmedia/open-api-framework/pull/31 to be merged and released.

Coperh:
Waiting for Other PRs to merge

src/openklant/accounts/tests/test_admin.py
```diff
@@ -0,0 +1,45 @@
...
+
+LOGIN_URL = reverse_lazy("admin:login")
+
+
+@override_settings(
+    USE_OIDC_FOR_ADMIN_LOGIN=False,
+    MAYKIN_2FA_ALLOW_MFA_BYPASS_BACKENDS=[],  # enforce MFA
+)
+class RecoveryTokenTests(WebTest):
+    @tag("gh-4072")
```
Coperh, line 17: Do we care enough to remove this tag?

stevenbal: Hmm yeah, let's just remove it

stevenbal (changes requested):
Works as expected, could you also add `DISABLE_2FA` to the docker-compose envvars (with default `True`)?

I do notice that some templates show the admin version though, these should not be visible until the user is fully authenticated

after adding 2FA:
image
backuptokens template:
image

I think this can be fixed by overriding the base template footer: https://github.com/open-zaak/open-zaak/blob/main/src/openzaak/templates/maykin_2fa/base.html

stevenbal (approved):

PR #202, opened 2024-07-11
:construction_worker: add gh actions to check OAS

annashamray:
Fixes https://github.com/open-zaak/open-zaak/issues/1708 (Open Klant part)


src/openklant/components/contactgegevens/api/schema.py
```diff
@@ -9,4 +9,8 @@
     "TITLE": "contactgegevens",
     "DESCRIPTION": description,
     "VERSION": settings.CONTACTGEGEVENS_API_VERSION,
+    "TAGS": [
```
stevenbal, line 12: The schemas probably have to be regenerated to add these tags?

annashamray: Yes. Updated OAS

stevenbal (approved):

PR #206, opened 2024-07-23
📝 [open-zaak/open-zaak#1649] Document envvars

stevenbal:
Fixes https://github.com/open-zaak/open-zaak/issues/1649 partially

**Changes**

* Document the envvars
* Add CI job to check if envvar docs were updated

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,6 +102,31 @@ jobs:
       - name: Publish coverage report
         uses: codecov/codecov-action@v3
 
+  check-envvar-docs:
+    runs-on: ubuntu-latest
+    name: Documentation build
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+          cache: 'pip'
+          cache-dependency-path: 'requirements/*.txt'
+      - name: Install dependencies
+        run: pip install -r requirements/ci.txt pytest
+      - name: Generate environment variable documentation using OAf and check if it was updated
+        run: |
+          bin/generate_envvar_docs.sh
+          changes=$(git diff docs/installation/config.rst)
+          if [ ! -z "$changes" ]; then
+              echo $changes
+              echo "Please update the environment documentation by running \`bin/generate_envvar_docs.sh\`"
+              exit 1
+          fi
+        env:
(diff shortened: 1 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -102,6 +102,31 @@ jobs:
...
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+          cache: 'pip'
+          cache-dependency-path: 'requirements/*.txt'
+      - name: Install dependencies
+        run: pip install -r requirements/ci.txt pytest
+      - name: Generate environment variable documentation using OAf and check if it was updated
+        run: |
+          python src/manage.py generate_envvar_docs --file docs/installation/config.rst --exclude-group Celery
```
stevenbal, line 120: Excluded celery settings here, it does seem to be installed, but I don't think it's actually used? @bart-maykin do you know more about this?

annashamray, line 120: Let's have a bash file with it, it can be useful for generating docs
Too bad it seems impossible to add it as sphinx event

annashamray (changes requested):

.github/workflows/ci.yml
```diff
@@ -102,6 +102,31 @@ jobs:
       - name: Publish coverage report
         uses: codecov/codecov-action@v3
 
+  docs:
+    runs-on: ubuntu-latest
+    name: Documentation build
```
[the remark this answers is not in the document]
stevenbal: I guess this will be tackled in https://github.com/maykinmedia/open-api-framework/issues/13 already?

.github/workflows/ci.yml
stevenbal: It could be, but I'm not sure how useful it is, since different projects will have different (or no) values set for `--exclude-group` and might also have a different target file and in that case the script would just be an alias for `python src/manage.py generate_envvar_docs`

I'll add one for each project that has the correct flags use that in CI, to make it easy to locally run the same command as in CI with `bin/generate_envvar_docs.sh`

.github/workflows/ci.yml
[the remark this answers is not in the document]
annashamray: Yes, sounds good

annashamray (approved):
Looks nice and neat

PR #218, opened 2024-08-08
:sparkles: [#182] changed internetaken actor field to actoren

### LINKED ISSUE #182: Taak kunnen toewijzen aan medewerker én een organisatorische eenheid
### Thema / Theme

Klantinteracties API

Als werkverdeler binnen de gemeente
wil ik dat een Interne Taak naast een medewerker ook binnen een specifieke afdeling (organisatorische eenheid) wordt toegewezen
zodat ik deze taak kan zien binnen het geheel van werkzaamheden van mijn team en kan toewijzen aan een beschikbare collega
 

_No response_

Voor details zie https://github.com/VNG-Realisatie/klantinteracties/issues/284

bart-maykin:
Fixes #182

**Changes**

- changed the actor fk field in internetaken to a m2m field named actoren.
- changed the api spec to support support the m2m field whilst supporting the old behavior (with the added deprecated label)
- created a data migration to transfer the data off the fk field to the m2m field


src/openklant/components/klantinteracties/migrations/0016_remove_internetaak_actor_interneactorenthoughmodel_and_more.py
```diff
@@ -0,0 +1,70 @@
...
+            model_name="internetaak",
+            name="actoren",
+            field=models.ManyToManyField(
+                help_text="De actoren aan wie de interne taak werd toegewezen.",
+                through="klantinteracties.InterneActorenThoughModel",
+                to="klantinteracties.actor",
+                verbose_name="actoren",
+            ),
+        ),
+        migrations.RunPython(sync_actor_and_actoren),
```
annashamray, line 65: please add also data migration for reverse process

src/openklant/components/klantinteracties/api/serializers/internetaken.py
```diff
@@ -33,12 +38,21 @@ class Meta:
         }
 
 
+@extend_schema_serializer(deprecate_fields=["toegewezen_aan_actor"])
 class InterneTaakSerializer(serializers.HyperlinkedModelSerializer):
     toegewezen_aan_actor = ActorForeignKeySerializer(
-        required=True,
+        required=False,
```
annashamray, line 44: Now it's possible that a task doesn't have any actor, isn't it? 
Is this change intended? Can it affect a process in some way?

src/openklant/components/klantinteracties/models/internetaken.py
```diff
@@ -98,3 +99,15 @@ def save(self, *args, **kwargs):
 
     def __str__(self):
         return f"{self.klantcontact} - ({self.nummer})"
+
+
+# Added for deprecated toegewezen_aan_actor field to return correct order
+class InterneActorenThoughModel(OrderedModel):
```
annashamray, line 105:
```suggestion
class InterneTakenActorenThoughModel(OrderedModel):
```

src/openklant/components/klantinteracties/api/serializers/internetaken.py
```diff
@@ -70,7 +85,52 @@ class Meta:
...
+            instance.actoren.order_by("interneactorenthoughmodel__order").first(),
+            context={**self.context},
+        ).data
+
+        response["toegewezen_aan_actoren"] = ActorForeignKeySerializer(
+            instance.actoren.all().order_by("interneactorenthoughmodel__order"),
+            context={**self.context},
+            many=True,
+        ).data
+        return response
```
annashamray, line 100: Is this code necessary? It looks like it duplicates the code with field declaration

src/openklant/components/klantinteracties/api/serializers/internetaken.py
```diff
@@ -70,7 +85,52 @@ class Meta:
...
+
+    # TODO: remove when depricated actoren field gets removed
+    def _validate_actoren(self):
+        toegewezen_aan_actor = "toegewezen_aan_actor" in self.initial_data
+        toegewezen_aan_actoren = "toegewezen_aan_actoren" in self.initial_data
+
+        if toegewezen_aan_actor == toegewezen_aan_actoren:
```
annashamray, line 107: Is it possible? They should have different types since one is iterable and another one is not
What happens if they are not equal? Could you please add a test for it

src/openklant/components/klantinteracties/api/serializers/internetaken.py
```diff
@@ -70,7 +85,52 @@ class Meta:
...
+                    return
+
+            raise serializers.ValidationError(message)
+
+    # TODO: remove when depricated actoren field gets removed
+    def _get_actoren(self, actoren):
+        if isinstance(actoren, list):
+            actor_uuids = [str(actor.get("uuid")) for actor in actoren]
+        else:
+            actor_uuids = [glom(actoren, "first.uuid", skip_exc=PathAccessError)]
```
annashamray, line 128: I don't quite understand this line. Could you please provide an example when it happens?

src/openklant/components/klantinteracties/api/serializers/internetaken.py
```diff
@@ -89,21 +149,33 @@ def validate(self, attrs):
...
+        return internetaak
 
     @transaction.atomic
     def update(self, instance, validated_data):
-        if "actor" in validated_data:
-            if actor := validated_data.pop("actor", None):
-                validated_data["actor"] = Actor.objects.get(uuid=str(actor.get("uuid")))
+        if "actoren" in validated_data:
+            actoren = validated_data.pop("actoren")
+            instance.actoren.clear()
```
annashamray, line 173: Always delete and create relations seems unnecessary. I think it's better to compare first that old and new actors are not the same

annashamray (dismissed):
I've looked through the code, but I don't have enough domain knowledge about this API to understand if the it's what the client wants. For this goal it's better to wait for Joeri or somebody else know knows Open Klant API

src/openklant/components/klantinteracties/api/serializers/internetaken.py
bart-maykin: It is indeed not intended for a `IntereTaak` to not have an `Actor`.
But because we now have 2 (1 temporary) `Actor` fields we can't have them set as required.
So I moved the "required" clause to the validation function within the serializer.

src/openklant/components/klantinteracties/api/serializers/internetaken.py
bart-maykin: It unfortunately is necessary. The issue is very well described in the Django Ordered Model library:
https://github.com/django-ordered-model/django-ordered-model?tab=readme-ov-file#ordering-of-manytomany-relationship-query-results

src/openklant/components/klantinteracties/api/serializers/internetaken.py
bart-maykin: I am basically checking here if the fields got send by the user.
If `toegewezen_aan_actor` is True then the user send it, if not then it is False and the same goes for `toegewezen_aan_actoren`.

This way I can validate if only one of the two fields got used. If both are False then neither were send, and if both are True then both got send.

src/openklant/components/klantinteracties/api/serializers/internetaken.py
bart-maykin: The structure of both actor fields are different.

The single object returns a validated data of {"first": {"uuid": "..."}}
And the multiple objects return a validated data of [{"uuid": "..."}, {"uuid": "..."}]

This is just a temporary function to fetch the data I need to update and create the objects.

src/openklant/components/klantinteracties/api/serializers/internetaken.py
bart-maykin: This will be done in the future but because we currently are bound to a temporary ordering model. I think it is not worth the time and resources to create the functionality of comparing the data between the existing data and the user submitted data and compare the order of both.

Deleting the relations and creating them afterwards will guarantee that the order will always be correct. So I think that for now this is a fine solution.

alextreme:
Besproken met Bart, en de feedback van Anna en de twijfels bij Bart laten zien dat er hier (in hun en mijn ogen onnodige) complexiteit wordt geintroduceerd. Bovendien is in de refinement te weinig aandacht geweest voor het ordering binnen de reeks van actoren die hierdoor ontstaat.

Deze PR gaat onhold voor nu totdat Joeri terug is, de specificatie is te weinig uitgewerkt (en spreekt zichzelf tevens tegen) om hier een klap op te geven

src/openklant/components/klantinteracties/api/serializers/internetaken.py
annashamray: Ok, let's keep it until we get rid of the deprecated field and OrderedModel

joeribekker:
1. Write `toegewezenAanActor=A`
2. Read: `toegewezenAanActor=A`, `toegewezenAanActoren=[A]`

1. Write `toegewezenAanActor=null`
2. Read: `toegewezenAanActor=null`, `toegewezenAanActoren=[]`

1. Write `toegewezenAanActor=B` (was A)
2. Read: `toegewezenAanActor=B`, `toegewezenAanActoren=[B]`

1. Write `toegewezenAanActor=B` (toegewezenAanActoren was A, B)
2. Read: `toegewezenAanActor=B`, `toegewezenAanActoren=[B]`

1. Write `toegewezenAanActoren=[A, B]`
2. Read: `toegewezenAanActor=A`, `toegewezenAanActoren=[A, B]`

1. Write `toegewezenAanActoren=[B, A]`  (was A, B)
2. Read: `toegewezenAanActor=B`, `toegewezenAanActoren=[B, A]`

src/openklant/components/klantinteracties/admin/internetaken.py
```diff
@@ -1,5 +1,7 @@
 from django.contrib import admin
 
+from ordered_model.admin import OrderedInlineModelAdminMixin
```
joeribekker, line 3: This seems unneeded now?

joeribekker (approved):

src/openklant/components/klantinteracties/api/serializers/internetaken.py
annashamray: Is it possible to modify it now?

annashamray (commented):
Looks good, but it's compare old and new actoren before removing them in update action

PR #223, opened 2024-08-15
:wrench: [maykinmedia/open-api-framework#40] Remove sites dependency

stevenbal:
Fixes https://github.com/maykinmedia/open-api-framework/issues/40 partially

**Changes**

* Uninstall the django.contrib.sites app to ensure that we no longer rely on in when generating the label for 2FA

### DIFF
--- a/src/openklant/conf/base.py
+++ b/src/openklant/conf/base.py
@@ -15,6 +15,9 @@ INSTALLED_APPS = INSTALLED_APPS + [
     "openklant.components.klantinteracties",
     "openklant.components.contactgegevens",
 ]
+# `django.contrib.sites` is installed by Open API Framework by default
+# but we don't want to rely on it anymore (e.g. when generating the label for 2FA)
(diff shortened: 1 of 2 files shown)


annashamray (commented):
So you decided to keep `sites` in OAF and remove it in APIs themselves?

Afaik all our API admins have 2FA now, therefore they all suffer from the same problem. 
We include 2fa in OAF, so it would be logical to fix the problem and removed `Sites` from OAF settings, I think

annashamray:
Btw I haven't noticed notification sending in Open Klant so I think it would be safe to remove Sites here

stevenbal:
@annashamray from the discussion we had on tuesday about this, I think removing sites per project (where possible) is the way to go for now right? https://github.com/maykinmedia/open-api-framework/issues/40#issuecomment-2298361383

annashamray:
Yes, let's for it this way

annashamray (approved):

PR #255, opened 2024-09-27
:wrench:[maykinmedia/open-api-framework#68] fix CSP errors

Coperh:
Fixes maykinmedia/open-api-framework#68


.github/workflows/oaf-check.yml
```diff
@@ -0,0 +1,17 @@
+name: Check Open API Framework Version
```
annashamray, line 1: It looks like this CI check is not in the scope of the bugfix

.github/workflows/oaf-check.yml
```diff
@@ -0,0 +1,17 @@
...
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ "master" ]
```
annashamray, line 8: afair we've decided to include all branches for the CI
But it could be moved to the separate PR

Coperh: Yeah, this is an old test version. The up-to-date version is in objects API

stevenbal (approved):

Issue #197, opened 2024-06-19
Develop OK v1 to OKv2 Klanten migration script


alextreme:
### Thema / Theme

Other

### Omschrijving / Description

Taiga DH 579

A script to migrate the Open Klant v1 Klanten to Open Klant v2 has been requested, for further details of which fields are to be migrated see the Taiga issue

joeribekker:
Estimated on 5 days

joeribekker:
@JanBrek ping

PeterVanBragt:
At first glance, 5 days seems like a lot of work for a query. Can some more background information be provided on this?

joeribekker:
@PeterVanBragt it's 2 different databases, so just a query won't cut it. Let's not downplay the work involved here.

However, if the scope is limited to "klanten" to be migrated to "partijen" (explicitly saved) and leave out "klanten" (ad hoc saved), things will be little easier since we don't need much logic. The tables however are really different, so we need to export it and then when importing, scatter the data over various tables.

There was a reason Open Klant had a major version bump. The data is not really compatible.

So. Maybe 4 days if no logic do distinct between partijen and klanten.

PeterVanBragt:
Hi @joeribekker, I can hereby approve this issue!

alextreme:
Discussed, migration script will be rounded up and an Open Klant release with it will be ready before the end of november

PR #263, opened 2024-10-11
[#197] Klanten API migration command

SonnyBA:
Fixes [#197](https://github.com/maykinmedia/open-klant/issues/197)

**Changes**

Adds a management command which allows the migration of the previous Klanten API to the Klantinteracties API. Two running instances (Klanten and Klantinteracties API) are required to do so with an API token for the Klanten API.


src/openklant/management/commands/migrate_to_v2.py
```diff
@@ -0,0 +1,475 @@
...
+        url_validator = URLValidator(schemes=["http", "https"])
+
+        for url in (v1_url, v2_url):
+            try:
+                url_validator(url)
+            except ValidationError as e:
+                return (
+                    f"Invalid URL(s) detected: {str(e.message)}. See "
+                    "migrate_to_v2 --help for correct usage."
+                )
```
stevenbal, line 460: We could maybe raise a `CommandError` here to make the command return an error code as well, but it's a minor thing

src/openklant/management/commands/migrate_to_v2.py
```diff
@@ -0,0 +1,475 @@
...
+        return Partij(**data)
+
+
+def _retrieve_klanten(url: str, access_token: str) -> Tuple[list[Klant], str | None]:
+    klanten_path = "/klanten/api/v1/klanten"
+
+    _url = urlparse(url)
+    _params = parse_qs(_url.query)
+
+    client = LegacyOpenKlantClient(f"{_url.scheme}://{_url.netloc}", access_token)
```
stevenbal, line 338: We could use `furl` here to construct the URL and define the client outside of the `_retrieve_klanten` function and pass that as an argument instead, to avoid having to do it multiple times, but I suppose it's not too much of an issue

src/openklant/management/commands/migrate_to_v2.py
```diff
@@ -0,0 +1,475 @@
...
+    response_data = client.retrieve(_url.path or klanten_path, params=_params)
```
stevenbal, line 339: Is it necessary to pass the `params` in separately? Can it not be part of the first argument?

src/openklant/management/commands/migrate_to_v2.py
```diff
@@ -0,0 +1,475 @@
...
+
+    if not isinstance(response_data, dict):
+        logger.error(f"Unexpected response data returned: {response_data}")
+        return [], None
+
+    items = response_data.get("results", [])
+    klant_fields = {field.name: field for field in dataclass_fields(Klant)}
```
stevenbal, line 346: This could also be defined outside of this function I think (perhaps as a constant)

src/openklant/management/commands/migrate_to_v2.py
```diff
@@ -0,0 +1,475 @@
...
+
+    digitaal_addressen_path = reverse(
+        "klantinteracties:digitaaladres-list",
+        kwargs=dict(version=settings.REST_FRAMEWORK["DEFAULT_VERSION"]),
+    )
+
+    partijen_path = reverse(
+        "klantinteracties:partij-list",
+        kwargs=dict(version=settings.REST_FRAMEWORK["DEFAULT_VERSION"]),
+    )
```
stevenbal, line 375: These can also be defined as constants outside of the function I think

src/openklant/management/commands/migrate_to_v2.py
```diff
@@ -0,0 +1,475 @@
...
+            _data = openklant_client.create(
+                digitaal_addressen_path, digitaal_adres.dict()
+            )
+
+            if not isinstance(_data, dict):
+                logger.error(
+                    f"Unknown data received creating a digitaal adres: {_data}. "
+                    "Skipping klant."
+                )
+                continue
```
stevenbal, line 396: I'm not sure what the user can actually do if some entries fail, but will it be visible in the logs which specific Klanten have failed so that they can be easily identified?

src/openklant/management/commands/migrate_to_v2.py
```diff
@@ -0,0 +1,475 @@
...
+
+        results = []
+
+        while next_url is not None:
+            klanten, next_url = _retrieve_klanten(
+                v1_url if next_url == "" else next_url, access_token
+            )
+            results.extend(_save_klanten(v2_url, klanten))
+
+        return "\n".join(results)
```
stevenbal, line 475: I see this returns the results, but since it's a `Command`, is this also part of the output you see when running the command?

src/openklant/tests/fixtures/migrate_command/cassettes/digitaal_adres.yaml
```diff
@@ -0,0 +1,47 @@
+interactions:
```
stevenbal, line 1: I think I would put these cassettes under `src/openklants/tests/files/vcr_cassettes/...` instead of `/fixtures`

stevenbal (changes requested):

stevenbal:
@SonnyBA I haven't worked on this yet, since I've spent most of my time on zgw consumers upgrade for Open Zaak. Could you pick up the feedback for this PR?

joeribekker:
Refinement: Waiting for @SonnyBA to get better. If not this week, @annashamray will pick it up.

src/openklant/tests/fixtures/migrate_command/cassettes/digitaal_adres.yaml
SonnyBA: I put these in the `migration/cassettes` folder, where there also is an fixture folder (for the V1 setup).

SonnyBA:
Requires https://github.com/maykinmedia/open-klant/pull/280

src/openklant/management/commands/migrate_to_v2.py
SonnyBA: Instantiating the client outside of the `_retrieve_klanten` function would require parsing the query parameters whenever multiple pages exist (pagination) and passing that to the function. Either way we'll have to parse the given URL somewhere.

src/openklant/management/commands/migrate_to_v2.py
SonnyBA: We could make it part of the first argument but it would require some extra work, while this currently reuses the `_params` and uses the `params` keyword available from `requests` (and does the work for us).

SonnyBA:
@stevenbal I moved code from the management command file to the `migration` folder inside the django project dir, this makes things more readable in my opinion (separate folders for `v1`/`v2` definitions) instead of having everything in one file. I also moved the cassettes to the `migration` folder inside of the project root where there also is a `fixtures` directory (which is used for the docker compose setup).

alextreme:
@SonnyBA please round up this PR and release a new version of Open Klant with in

docs/development/migration.rst
```diff
@@ -0,0 +1,49 @@
+.. _migration_index:
+
+Migration
+=========
+
+Users of Open Klant version ``1.0.0`` willing to migrate to version ``2.0.0`` or higher
+can make use of the ``migrate_to_v2`` management command. This command **only** migrates
+``Klant`` instances from version ``1.0.0``. The command expects two URLs; a URL for the
+Open Klant instance of version ``1.0.0.``and a URL for version ``2.0.0``. The command
```
stevenbal, line 9:
```suggestion
Open Klant instance of version ``1.0.0`` and a URL for version ``2.0.0`` The command
```

docs/development/migration.rst
```diff
@@ -0,0 +1,49 @@
+.. _migration_index:
+
+Migration
```
stevenbal, line 3: I initially read this as documentation for devops, but looking at the lack of docs for Open Klant, I don't think any client/devops will find these docs.

Do we need to document/mention this command in the changelog to make sure at least Sjoerd knows about it?

docs/development/migration.rst
```diff
@@ -0,0 +1,49 @@
...
+
+    # Start the docker compose setup
+    $ docker compose up --detach
+
+    # Load a fixture for a certain test
+    $ docker compose exec -it web \
+        ./src/manage.py loaddata migration/credentials migration/test_digitaal_adres
+
+    # Run the corresponding test
+    $ ./src/manage.py test test openklant.tests.test_migrate.MigrateTestCase.test_digitaal_adres
```
stevenbal, line 35:
```suggestion
    $ ./src/manage.py test openklant.tests.test_migrate.MigrateTestCase.test_digitaal_adres
```

src/openklant/management/commands/migrate_to_v2.py
```diff
@@ -0,0 +1,176 @@
...
+        response_data = openklant_client.create(PARTIJEN_PATH, partij.dict())
+
+        if response_data and "url" in response_data:
+            created_klanten.append(response_data["url"])
+
+    return created_klanten
+
+
+def _generate_dummy_token() -> str:
+    token_auth, _ = TokenAuth.objects.get_or_create(application="Migration application")
```
stevenbal, line 127: It might be good to remove this token after running the command (and also if the command fails)

src/openklant/management/commands/migrate_to_v2.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+def _save_klanten(url: str, klanten: list[Klant]) -> list[str]:
+    if not klanten:
+        logger.info("No klanten to save, moving on...")
+        return []
+
+    logger.info(f"Trying to create {len(klanten)} klanten through the V2 API")
+
+    dummy_token = _generate_dummy_token()
```
stevenbal, line 90: If I understand this correctly, it if the list is paginated and `_save_klanten` is run multiple times, you'll have multiple dummy tokens. It might be better to call this outside the loop in `.handle` to avoid this.

stevenbal (changes requested):
Some small questions/remarks, otherwise it looks good :+1:

src/openklant/management/commands/migrate_to_v2.py
SonnyBA: While it used `get_or_create` to retrieve the token before changes were made to this, I agree that querying the database multiple times is unnecessary, see https://github.com/maykinmedia/open-klant/pull/263/commits/293156ee77fa8ee876d8d0d92053eaf9e6cd1c7f.

src/openklant/management/commands/migrate_to_v2.py
SonnyBA: See https://github.com/maykinmedia/open-klant/pull/263/commits/293156ee77fa8ee876d8d0d92053eaf9e6cd1c7f

docs/development/migration.rst
SonnyBA: I moved the generic part of this file to the user targeted documentation (under `installation`) and also updated the changelog.

docs/installation/migration.rst
```diff
@@ -0,0 +1,21 @@
...
+``Klant`` instances from version ``1.0.0``. The command expects two URLs; a URL for the
+Open Klant instance of version ``1.0.0`` and a URL for version ``2.0.0``.
+
+The command also expects a ``ACCESS_TOKEN`` environment variable to be set, to authenticate to
+the Open Klant ``1.0.0`` instance. For the ``2.0.0`` instance a dummy token will be
+created. This dummy token will be removed after the command ran (be it successfully or not).
+
+An example of how one might want to run this command can be seen below:
+
+```
```
stevenbal, line 17: formatting is off for this:

image

you might have to use `.. code-block:: bash`

stevenbal (approved):
Only one remark about docs formatting, otherwise good to merge for me :+1:

docs/installation/migration.rst
SonnyBA: Good catch! That fixed it:

image

PR #271, opened 2024-10-18
 ✨ [#234] Validate DigitaalAdres.adres if type is email

### LINKED ISSUE #234: Validation of any stored email address
### Thema / Theme

Klantinteracties API

It is desirable to have OpenKlant validate any email addresses before storing them so that they are stored in a consistent way.

This make sure only valid mail addresses are stored and any connected system can rely on the provided addresses.

As discussed on 29-8 with Andy, ICATT & Maykin

stevenbal:
Fixes #234 

**Changes**

*  Validate DigitaalAdres.adres if type is email

### DIFF
--- a/src/openklant/components/klantinteracties/api/serializers/digitaal_adres.py
+++ b/src/openklant/components/klantinteracties/api/serializers/digitaal_adres.py
@@ -6,7 +6,10 @@ from rest_framework import serializers
 from openklant.components.klantinteracties.api.serializers.constants import (
     SERIALIZER_PATH,
 )
-from openklant.components.klantinteracties.api.validators import digitaal_adres_exists
+from openklant.components.klantinteracties.api.validators import (
+    OptionalEmailValidator,
+    digitaal_adres_exists,
+)
 from openklant.components.klantinteracties.models.digitaal_adres import DigitaalAdres
 from openklant.components.klantinteracties.models.klantcontacten import Betrokkene
 from openklant.components.klantinteracties.models.partijen import Partij
@@ -85,6 +88,7 @@ class DigitaalAdresSerializer(serializers.HyperlinkedModelSerializer):
                 "help_text": _("De unieke URL van dit digitaal adres binnen deze API."),
             },
         }
+        validators = [OptionalEmailValidator()]
 
     @transaction.atomic
     def update(self, instance, validated_data):
--- a/src/openklant/components/klantinteracties/api/serializers/digitaal_adres.py
+++ b/src/openklant/components/klantinteracties/api/serializers/digitaal_adres.py
@@ -13,6 +13,7 @@ from openklant.components.klantinteracties.api.validators import (
 from openklant.components.klantinteracties.models.digitaal_adres import DigitaalAdres
 from openklant.components.klantinteracties.models.klantcontacten import Betrokkene
 from openklant.components.klantinteracties.models.partijen import Partij
+from openklant.utils.serializers import get_field_value
 
 
 class DigitaalAdresForeignKeySerializer(serializers.HyperlinkedModelSerializer):
@@ -88,7 +89,17 @@ class DigitaalAdresSerializer(serializers.HyperlinkedModelSerializer):
                 "help_text": _("De unieke URL van dit digitaal adres binnen deze API."),
             },
         }
-        validators = [OptionalEmailValidator()]
+
+    def validate_adres(self, adres):
+        """
+        Define the validator here, to avoid DRF spectacular marking the format for
+        `adres` as `email`
+        """
+        soort_digitaal_adres = get_field_value(
+            self, self.initial_data, "soort_digitaal_adres"
+        )
+        OptionalEmailValidator()(adres, soort_digitaal_adres)
+        return adres
 
     @transaction.atomic
     def update(self, instance, validated_data):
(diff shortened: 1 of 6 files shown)


src/openklant/components/klantinteracties/api/tests/test_digitaal_adres.py
```diff
@@ -90,6 +92,54 @@ def test_create_digitaal_adres(self):
             self.assertEqual(data["adres"], "foobar@example.com")
             self.assertEqual(data["omschrijving"], "omschrijving")
 
+    def test_create_digitaal_adres_email_validation(self):
```
SonnyBA, line 95: Could you also add a test which verifies that whenever `SoortPartij` is not `email` no email validation is done?

src/openklant/components/klantinteracties/api/serializers/digitaal_adres.py
```diff
@@ -85,6 +88,7 @@ class Meta:
                 "help_text": _("De unieke URL van dit digitaal adres binnen deze API."),
             },
         }
+        validators = [OptionalEmailValidator()]
```
SonnyBA, line 91: It would be nice if this validation is also done in the admin (not only through the API).

SonnyBA (dismissed):

src/openklant/components/klantinteracties/api/tests/test_digitaal_adres.py
stevenbal: This is already tested as part of `test_update_digitaal_adres` which uses `SoortDigitaalAdres.telefoonnummer`, is that enough or should I add a separate test case for it?

src/openklant/components/klantinteracties/api/serializers/digitaal_adres.py
stevenbal: I've added this in https://github.com/maykinmedia/open-klant/pull/271/commits/90076edd6199e6c6c79d42260233aa926581d9c5

src/openklant/components/klantinteracties/api/validators.py
```diff
@@ -166,3 +170,26 @@ def Rekeningnummer_exists(value):
...
+            == SoortDigitaalAdres.email
+        ):
+            try:
+                super().__call__(get_field_value(serializer, attrs, "adres"))
+            except ValidationError as e:
+                # re-raise as field error
+                if serializer:
+                    raise exceptions.ValidationError({"adres": e.message})
+                else:
+                    raise ValidationError({"adres": e})
```
stevenbal, line 195: Unfortunately I found no way around this, ideally I'd always raise a `django.core.exceptions.ValidationError`, but I wasn't able to raise that as a field specific error in rest_framework (it was always in the data as `nonFieldErrors`)

SonnyBA: Do you mean it shows up as `nonFieldErrors` in the admin? It might be related to calling the validator in the `clean` [method](https://github.com/maykinmedia/open-klant/pull/271/files#diff-4086aff50bf30b40ef88ec2f2299d06c4c266ae0e908cbe02c765e54f1079cacR13) (as opposed to the `clean_<field>`)

stevenbal: No in the admin it works as expected, but the API returns it as a nonfielderror, regardless of whether I re-raise the error like this: `raise ValidationError({"adres": e})`

SonnyBA: hmmm, what about not using a validator (at the serializer level) but using the `validate` method on the specific serializer (as this validation is done for a very specific use case)?

stevenbal: You're right, I've changed it so now I do not have to re-raise the error anymore

src/openklant/components/klantinteracties/api/validators.py
```diff
@@ -166,3 +171,27 @@ def Rekeningnummer_exists(value):
...
+class OptionalEmailValidator(EmailValidator):
+    """
+    EmailValidator for SoortDigitaalAdres that only attempts to validate if
+    `SoortDigitaalAdres` is `email`
+    """
+
+    requires_context = True
+
+    def __call__(
+        self, value: str, serializer_field: serializers.Field, attrs: dict | None = None
```
annashamray, line 185:
```suggestion
        self, value: str, serializer_field: serializers.Field | None, attrs: dict | None = None
```

src/openklant/components/klantinteracties/api/validators.py
```diff
@@ -166,3 +171,27 @@ def Rekeningnummer_exists(value):
...
+    ):
+        if serializer_field:
+            serializer = serializer_field.parent
+            serializer = cast(serializers.Serializer, serializer)
+            soort_digitaal_adres = get_field_value(
+                serializer, serializer.initial_data, "soort_digitaal_adres"
+            )
+        else:
+            soort_digitaal_adres = attrs.get("soort_digitaal_adres")
```
annashamray, line 194: A little nitpicking: this part doesn't actually provide logic for OptinalEmailValidator, it's more about how to retrieve `soort_digitaal_adres`. I'd move this part to serializer.validate_adres and form clean_adres methods.
And the validator itself would have parameters: 
```
    def __call__(self, value: str,  soort_digitaal_adres:str):
```

But it really is a matter of preference so please treat my comment as optional

annashamray (approved):
Looks good!

src/openklant/components/klantinteracties/api/validators.py
stevenbal: You're right, initially I added the validator to `validators` on the serializer.Meta, that's why I had this setup here. But since I manually call the validator now in `.validate`, it makes more sense to do it like you suggested

PR #273, opened 2024-10-22
:bug:[#226] make digitalAdres betrokkene optional in admin

### LINKED ISSUE #226: Question: why is `betrokkene` required in the Admin Interface, for `digitaalAdres` that is `verstrektDoorPartij`

2.1.0

I want to enter a Digitaal Adres for one of the Partijen we created from KISS, in the admin interface of OpenKlant. 

But when I try to add a Digitaal Adres for only a Partij, I get a message that `betrokkene` is required. 

From the OpenAPI.yaml, i got the impression that within `digitaalAdres` both `vertrektDoorBetrokkene` and `vertrektDoorPartij` are required, but can be null. Where the intention is: one of these two must not be null

What is the reason `betrokkene` is required when adding a `digitaalAdres`?
Is the idea that when a `partij` "verstrekt" a `digitaalAdres`, this is alway done within the context of a `klantcontact`, and hence there must be a `betrokkene` ?

1 Go  

_No response_

Coperh:
Fixes #226


src/openklant/components/klantinteracties/tests/test_admin.py
```diff
@@ -90,3 +94,35 @@ def test_search(self):
...
+        PersoonFactory(partij=partij)
+        url = reverse("admin:klantinteracties_partij_change", args=[partij.pk])
+
+        response = self.app.get(url)
+
+        form = response.form
+        form["digitaaladres_set-TOTAL_FORMS"] = 1
+        add_dynamic_field(form, "digitaaladres_set-0-omschrijving", "description")
+        add_dynamic_field(form, "digitaaladres_set-0-soort_digitaal_adres", "email")
+        add_dynamic_field(form, "digitaaladres_set-0-adres", "email@example.com")
```
stevenbal, line 119: As an alternative to having to add these fields, you could maybe create a DigitaalAdres with a factory to ensure the form is there, and then try to save it again without a betrokkene, but either way is fine to me

PR #284, opened 2024-11-19
:art: [#235] improve the way we validate phonenumbers

### LINKED ISSUE #235: Validation of any stored phone number
### Thema / Theme

Klantinteracties API

It is desirable to have OpenKlant validate and/or format/transform phone numbers before storing them so that they are stored in a consistent way.

This make sure only valid phone numbers are stored and any connected system can rely on the provided addresses.

As discussed on 29-8 with Andy, ICATT & Maykin

bart-maykin:
Fixes #235

**Changes**

Improved the way we validate phonenumbers with the phonenumbers library

### DIFF
--- a/src/openklant/components/klantinteracties/api/validators.py
+++ b/src/openklant/components/klantinteracties/api/validators.py
@@ -25,6 +25,7 @@ from openklant.components.klantinteracties.models.partijen import (
     PartijIdentificator,
 )
 from openklant.components.klantinteracties.models.rekeningnummers import Rekeningnummer
+from openklant.utils.validators import validate_phone_number
 
 
 class FKUniqueTogetherValidator(UniqueTogetherValidator):
@@ -170,12 +171,12 @@ def Rekeningnummer_exists(value):
         raise serializers.ValidationError(_("Rekeningnummer object bestaat niet."))
 
 
-class OptionalEmailValidator(EmailValidator):
-    """
-    EmailValidator for SoortDigitaalAdres that only attempts to validate if
-    `SoortDigitaalAdres` is `email`
-    """
-
-    def __call__(self, value: str, soort_digitaal_adres: str):
-        if soort_digitaal_adres == SoortDigitaalAdres.email:
-            super().__call__(value)
+class SoortDigitaalAdresValidator:
+    def __call__(self, soort_digitaal_adres: SoortDigitaalAdres, value: str):
+        match soort_digitaal_adres:
+            case SoortDigitaalAdres.email:
+                EmailValidator()(value)
+            case SoortDigitaalAdres.telefoonnummer:
+                validate_phone_number(value)
+            case _:
+                return
--- a/src/openklant/components/klantinteracties/tests/test_admin.py
+++ b/src/openklant/components/klantinteracties/tests/test_admin.py
@@ -123,11 +123,7 @@ class PartijAdminTests(WebTest):
 @disable_admin_mfa()
 class DigitaalAdresAdminTests(WebTest):
     @tag("gh-234")
-    def test_soort_digitaal_adres_validation(self):
-        """
-        Check that the admin applies conditional email/telefoonnummer validation on `adres`, only if
-        `soort_digitaal_adres` is `email` or 'telefoonnummer'
-        """
+    def test_email_validation(self):
         user = SuperUserFactory.create()
         self.app.set_user(user)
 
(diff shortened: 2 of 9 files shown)


src/openklant/utils/tests/test_validators.py
```diff
@@ -74,11 +74,20 @@ def test_validate_postal_code(self):
         self.assertIsNone(validate_postal_code("1015 cJ"))
 
     def test_validate_phone_number(self):
+        valid_phone_numbers = [
```
SonnyBA, line 77: Could you add the numbers here that are [mentioned here](https://github.com/maykinmedia/open-klant/issues/235#issuecomment-2323875962) (the top 5) to verify they work as expected?

src/openklant/utils/validators.py
```diff
@@ -24,12 +26,18 @@ def validate_charfield_entry(value, allow_apostrophe=False):
...
-        int(value.strip().lstrip("0+").replace("-", "").replace(" ", ""))
-    except (ValueError, TypeError):
+        phone_number = phonenumbers.parse(value, None)
+    except phonenumbers.NumberParseException:
         raise ValidationError(_("Het opgegeven mobiele telefoonnummer is ongeldig."))
 
-    return value
+    if not phonenumbers.is_possible_number(
+        phone_number
+    ) or not phonenumbers.is_valid_number(phone_number):
```
SonnyBA, line 39: Is `is_possible_number` needed when you call `is_valid_number`?

SonnyBA (changes requested):

src/openklant/utils/tests/test_validators.py
```diff
@@ -74,25 +74,37 @@ def test_validate_postal_code(self):
...
+            "+31612345678",
```
stevenbal, line 78: Would/should something like `0612345678` still be valid?

bart-maykin:
@stevenbal I replaced the `OptionalEmailValidator` with `SoortDigitaalAdresValidator`, because the original validator was specificly made anyway. I kept the same construction as the original implementation.

src/openklant/components/klantinteracties/api/validators.py
```diff
@@ -170,12 +171,21 @@ def Rekeningnummer_exists(value):
...
+    ):
+        self.email_validator = email_validator
+        self.phone_number_validator = phone_number_validator
+
+    def __call__(self, soort_digitaal_adres: SoortDigitaalAdres, value: str):
+        match soort_digitaal_adres:
+            case SoortDigitaalAdres.email:
+                self.email_validator(value)
+            case SoortDigitaalAdres.telefoonnummer:
+                self.phone_number_validator(value)
```
stevenbal, line 189: Minor thing, I'll approve but I think you could remove the init and do this:

```suggestion
            case SoortDigitaalAdres.email:
                EmailValidator()(value)
            case SoortDigitaalAdres.telefoonnummer:
                phone_number_validator(value)
```

stevenbal (approved):

src/openklant/utils/tests/test_validators.py
```diff
@@ -74,25 +74,38 @@ def test_validate_postal_code(self):
...
+                )
 
-        self.assertEqual(validate_phone_number(" 0695959595"), " 0695959595")
-        self.assertEqual(validate_phone_number("+33695959595"), "+33695959595")
-        self.assertEqual(validate_phone_number("00695959595"), "00695959595")
-        self.assertEqual(validate_phone_number("00-69-59-59-59-5"), "00-69-59-59-59-5")
-        self.assertEqual(validate_phone_number("00 69 59 59 59 5"), "00 69 59 59 59 5")
+        for valid_phone_number in valid_phone_numbers:
+            with self.subTest(valid_phone_number):
+                self.assertIsNone(validate_phone_number(valid_phone_number))
```
SonnyBA, line 108:
```suggestion
                validate_phone_number(valid_phone_number)
```
The test will raise an `ValidationError` when a number is not valid

src/openklant/components/klantinteracties/tests/test_admin.py
```diff
@@ -152,13 +152,33 @@ def test_email_validation(self):
                 response.context["errors"], [[_("Voer een geldig e-mailadres in.")]]
             )
 
-        with self.subTest("do not apply validation for soort_digitaal_adres != email"):
+        with self.subTest("apply validation for soort_digitaal_adres == email"):
```
SonnyBA, line 155:
```suggestion
        with self.subTest("apply validation for soort_digitaal_adres == telefoonnummer"):
```

src/openklant/components/klantinteracties/api/validators.py
SonnyBA: :point_up: , I don't think the `email_validator` and the `phone_number_validator` constructor keyword arguments are going to be used as `SoortDigitaalAdres` should use the same validation consistently

src/openklant/components/klantinteracties/tests/test_admin.py
```diff
@@ -123,10 +123,10 @@ def test_digitaal_adres_inline(self):
 @disable_admin_mfa()
 class DigitaalAdresAdminTests(WebTest):
     @tag("gh-234")
-    def test_email_validation(self):
+    def test_soort_digitaal_adres_validation(self):
```
SonnyBA, line 126: @bart-maykin could you split the three subtests into their own tests?

src/openklant/components/klantinteracties/api/validators.py
```diff
@@ -170,12 +171,12 @@ def Rekeningnummer_exists(value):
...
-class OptionalEmailValidator(EmailValidator):
-    """
-    EmailValidator for SoortDigitaalAdres that only attempts to validate if
-    `SoortDigitaalAdres` is `email`
-    """
-
-    def __call__(self, value: str, soort_digitaal_adres: str):
-        if soort_digitaal_adres == SoortDigitaalAdres.email:
-            super().__call__(value)
+class SoortDigitaalAdresValidator:
```
SonnyBA, line 174: This became a lot more readable  :+1:

SonnyBA (changes requested):
Almost there! Whenever those subtests are split up this can be merged if you ask me.

PR #305, opened 2024-12-12
[#293] update token configuration file namespaces

### LINKED ISSUE #293: Add support for Tokens via django-setup-configuration
### Thema / Theme

Klantinteracties API

Allow tokens to be configured.

_No response_

_No response_

SonnyBA:
Fixes #293 

**Changes**

Renames the configuration file namespaces for token configuration

### DIFF
--- a/docker/setup_configuration/data.yaml
+++ b/docker/setup_configuration/data.yaml
@@ -1,5 +1,5 @@
-openklant_tokens_config_enable: true
-openklant_tokens:
+zgw_tokens_tokenauth_config_enable: true
+zgw_tokens_tokenauth:
   items:
     - identifier: token-1
       token: ba9d233e95e04c4a8a661a27daffe7c9bd019067
--- a/docker/setup_configuration/data.yaml
+++ b/docker/setup_configuration/data.yaml
@@ -1,5 +1,5 @@
-zgw_tokens_tokenauth_config_enable: true
-zgw_tokens_tokenauth:
+tokenauth_config_enable: true
+tokenauth:
   items:
     - identifier: token-1
       token: ba9d233e95e04c4a8a661a27daffe7c9bd019067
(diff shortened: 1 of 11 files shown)


docker/setup_configuration/data.yaml
```diff
@@ -1,5 +1,5 @@
-openklant_tokens_config_enable: true
-openklant_tokens:
+zgw_tokens_tokenauth_config_enable: true
+zgw_tokens_tokenauth:
```
stevenbal, line 2: I thought we mentioned to call this `tokenauth_config_enable`? This prefix doesn't apply for Objects/Objecttypes, because those are not part of Zaakgericht werken

stevenbal (changes requested):

docker/setup_configuration/data.yaml
SonnyBA: @danielmursa-dev and I discussed that adding a prefix for the library that will be responsible for defining these tokens would be nice, because it takes away some work in advance as the configuration files will have the correct namespaces already, hence the `zgw_tokens` prefix. I guess the question here is, how are we going to call the library and use that prefix?

stevenbal: Hmm okay, I'll approve for now so that we can get the other PR in Objecttypes merged as well. We'll probably have to take a look at all the namespaces next week anyway

stevenbal (approved):

danielmursa-dev (approved):

docker/setup_configuration/data.yaml
stevenbal: I see that in https://github.com/maykinmedia/objecttypes-api/pull/135/files this was changed to just `tokenauth`, could you change it to that as well? I think no prefix is better than a prefix that might be confusing

PR #316, opened 2025-01-10
:wrench: Add timeouts to celery tasks

SilviaAmAm:
Done in the context of issue https://github.com/maykinmedia/charts/issues/148

**Changes**
- Adds a default soft and hard timeout to all celery tasks, to prevent cases where tasks hang and block a worker.


SonnyBA (approved):
Celery documentation [seems to be excluded ](https://github.com/maykinmedia/open-klant/blob/2d1abfa3163dfefe956c31a5bb827aea7c0aae69/bin/generate_envvar_docs.sh#L4) in this project when building documentation, looks good :+1:

PR #322, opened 2025-01-22
:fire: Remove unused command line args

SilviaAmAm:
Done as part of https://github.com/maykinmedia/charts/issues/165


stevenbal (approved):
As far as I know Open Klant doesn't use celery at the moment, but it doesn't hurt to change this (since support for notifications will probably be added at some point).

Issue #267, opened 2024-10-16
Enforce uniqueness of Partij


mariusvandam:
### Thema / Theme

Klantinteracties API

### Omschrijving / Description

It is currently possible to create multiple identical Partij Identificator records for the same 'Partij' 

As a replacement story for #241 we should create a more intelligent enforcement of uniqueness for partij. Depending on the used identificatoren for a Partij, different logic should be applied to make sure a single Partij is only created once.

1. Any `bsn` can only be occur once (natural persons)
2. A `kvkNummer` without a `vestigingsnummer` can only occur once (representing the "niet natuurlijk persoon" in general, separate from any concrete vestiging)
3. A combination of `kvkNummer` **AND** `vestigingsnummer` can only occur once (vestigingen, disambiguated according to their associated parent entity)
4. A `Partij` can **never** have a _vestigingsnummer_ assigned without a `kvkNummer` being assigned at the same time

This will almost certainly require some kind of bulk creation/deletion of identificatoren, to avoid to avoid intermediate states which would violate these invariants (e.g. if there's already a Partij with 1 `kvkNummer`, you would have to provide both `kvkNummer` and `vestigingsnummer` together for a second `Partij` you might want to create to represent a vestiging separately from the generic niet-natuurlijk persoon).

It would also require being able to filter the `Partij` resource on an array of identificator values (to filter a list down to a specific `kvk`/`vestigingsnummer` pairing (ideally using the identificator register/objecttype/id values directly, to avoid having to first fetch the partij identificatoren and filter on the identificator UUIDs). 

### Toegevoegde waarde / Added value

The value of an identificator should be that it actually functions as a unique identifier, both for developer experience (no difficult-to-resolve edge-cases to handle), but it would also improve the data quality (less chance of spurious duplicates). This would be complementary to https://github.com/maykinmedia/open-klant/issues/233 . Specifically it is important that there is no ambiguity when referring to vestigingen, where two vestigingen may have the same number but different parent `kvkNummer` entities.

joeribekker:
A `vestigingsnummer` is globally unique, so let's just do that instead of a combination of kvk number AND vestigingsnummer.

sytskevanhasselt:
See also [my comment](https://github.com/maykinmedia/open-klant/issues/233#issuecomment-2464562913) on issue #233

Hugo-ter-Doest:
@swrichards Yes, I think there can be only one PartijIdentificator per BSN.

[...]

PR #334, opened 2025-02-14
[#267] Enforce uniqueness of Partij

danielmursa-dev:
Fixes #


src/openklant/components/klantinteracties/models/validators.py
```diff
@@ -17,7 +19,147 @@
...
+
+        if self.queryset.filter(**filters).exists():
+            raise ValidationError(
+                {
+                    "__all__": _(
+                        "`PartijIdentificator` moet uniek zijn, er bestaat er al een met deze gegevenscombinatie."
+                    )
+                }
+            )
+        return
```
stevenbal, line 145:
```suggestion
```

src/openklant/components/klantinteracties/models/validators.py
```diff
@@ -49,18 +191,12 @@ class PartijIdentificatorValidator:
...
-        code_objecttype: str,
-        code_soort_object_id: str,
-        object_id: str,
-    ) -> None:
+    def __init__(self, partij_identificator: dict) -> None:
         """Initialize validator"""
-        self.code_register = code_register
-        self.code_objecttype = code_objecttype
-        self.code_soort_object_id = code_soort_object_id
-        self.object_id = object_id
```
stevenbal, line 63: Why is this change needed?

src/openklant/components/klantinteracties/tests/test_validators.py
```diff
@@ -330,10 +357,118 @@ def test_not_allowed_cases(self):
...
+                instance=partij_identificator,
+            ).validate()
+
+        details = error.exception.message_dict
+        self.assertEqual(
+            details["sub_identificator_van"][0],
+            "Kan zichzelf niet selecteren als `sub_identificator_van`.",
+        )
+
+    def test_invalid_sub_identificator_van_type_not_allowed(self):
```
stevenbal, line 429:
```suggestion
    def test_vestigingsnummer_requires_sub_identificator_van_with_kvk_nummer(self):
```

src/openklant/components/klantinteracties/tests/test_validators.py
```diff
@@ -330,10 +357,118 @@ def test_not_allowed_cases(self):
...
+        PartijIdentificatorUniquenessValidator(
+            partij_identificator={
+                "code_objecttype": PartijIdentificatorCodeObjectType.natuurlijk_persoon.value,
+                "code_soort_object_id": PartijIdentificatorCodeSoortObjectId.bsn.value,
+                "object_id": "296648875",
+                "code_register": PartijIdentificatorCodeRegister.brp.value,
+            },
+            sub_identificator_van=None,
+            instance=None,
+        ).validate()
```
stevenbal, line 383: can you also add an assertion here where you expect it to throw an error based on uniqueness?

src/openklant/components/klantinteracties/tests/test_validators.py
```diff
@@ -330,10 +357,118 @@ def test_not_allowed_cases(self):
...
+                },
+                sub_identificator_van=None,
+                instance=None,
+            ).validate()
+
+        details = error.exception.message_dict
+        self.assertTrue(
+            "`sub_identifier_van` met CodeSoortObjectId = `kvk_nummer` te kiezen."
+            in details["sub_identificator_van"][0]
+        )
```
stevenbal, line 474: could you make an assertion on the full message here instead?

src/openklant/components/klantinteracties/tests/test_validators.py
```diff
@@ -6,56 +6,70 @@
...
+)
 from openklant.components.klantinteracties.models.validators import (
-    PartijIdentificatorValidator,
+    PartijIdentificatorTypesValidator,
+    PartijIdentificatorUniquenessValidator,
 )
 
 
-class PartijIdentificatorValidatorTests(TestCase):
+class PartijIdentificatorTypesValidatorTests(TestCase):
```
stevenbal, line 18: This is missing a testcase for `validate_unique_partij_identificator_locally`, let's add one with `raise NotImplementedError` to make sure it's not forgotten

src/openklant/components/klantinteracties/api/tests/test_partijen.py
```diff
@@ -2067,158 +2080,164 @@ def test_invalid_choice_partij_identificator_code_objecttype(self):
...
-    def test_valid_validation_partij_identificator(self):
-        # All validations pass
-        url = reverse("klantinteracties:partijidentificator-list")
-        partij = PartijFactory.create()
+
+class PartijIdentificatorUniquenessTests(APITestCase):
+    def setUp(self):
+        self.list_url = reverse("klantinteracties:partijidentificator-list")
+        self.partij = PartijFactory.create()
+        super().setUp()
```
stevenbal, line 2236: `setUp` is run before every test in the class, we can use `reverse_lazy` at the class level to avoid running it more than necessary 
```suggestion
    list_url = reverse_lazy("klantinteracties:partijidentificator-list")
    
    def setUp(self):
        self.partij = PartijFactory.create()
        super().setUp()
```

src/openklant/components/klantinteracties/api/tests/test_partijen.py
```diff
@@ -2067,158 +2080,164 @@ def test_invalid_choice_partij_identificator_code_objecttype(self):
...
-            "voor `codeObjecttype` natuurlijk_persoon zijn alleen deze waarden toegestaan: ['bsn', 'overig']",
-        )
+        self.assertEqual(len(response.data["invalid_params"]), 4)
 
-    def test_invalid_validation_partij_identificator_object_id(self):
-        url = reverse("klantinteracties:partijidentificator-list")
+    def test_invalid_create_partial(self):
+        # all partij_identificator fields required
         partij = PartijFactory.create()
+        list_url = reverse("klantinteracties:partijidentificator-list")
```
stevenbal, line 2166: I notice that throughout this file this reverse call (and perhaps others as well) is duplicated a lot, can you move it to the class level with `reverse_lazy` instead?

src/openklant/components/klantinteracties/admin/partijen.py
```diff
@@ -18,6 +23,40 @@
...
+            "code_register": cleaned_data["partij_identificator_code_register"],
+        }
+
+        PartijIdentificatorTypesValidator(
+            partij_identificator=partij_identificator
+        ).validate()
+
+        queryset = PartijIdentificator.objects.exclude()
+        if self.instance:
+            queryset = queryset.exclude(pk=self.instance.pk)
```
stevenbal, line 48: This doesn't seem to be used anywhere?
```suggestion
```

src/openklant/components/klantinteracties/models/validators.py
```diff
@@ -17,7 +19,147 @@
...
+            filters["sub_identificator_van"] = self.sub_identificator_van
+
+        if self.queryset.filter(**filters).exists():
+            raise ValidationError(
+                {
+                    "__all__": _(
+                        "`PartijIdentificator` moet uniek zijn, er bestaat er al een met deze gegevenscombinatie."
+                    )
+                }
+            )
```
stevenbal, line 144: This means that the `invalid_params.name` in the API response will be `__all__`, while it should be `nonFieldErrors`

```suggestion
            raise ValidationError(
                _(
                        "`PartijIdentificator` moet uniek zijn, er bestaat er al een met deze gegevenscombinatie."
                 )
            )
```

stevenbal (changes requested):

src/openklant/components/klantinteracties/admin/partijen.py
danielmursa-dev: This is used when updating the instance and I want to exclude self from global uniqueness

src/openklant/components/klantinteracties/models/validators.py
danielmursa-dev: to be clear in `PartijIdentificatorGroepTypeSerializer`

```
    def validate(self, attrs):
        PartijIdentificatorTypesValidator(partij_identificator=attrs)()
        return super().validate(attrs)
```

src/openklant/components/klantinteracties/tests/test_validators.py
```diff
@@ -7,57 +7,62 @@
...
-            code_objecttype=PartijIdentificatorCodeObjectType.natuurlijk_persoon.value,
-            code_soort_object_id=PartijIdentificatorCodeSoortObjectId.bsn.value,
-            object_id="296648875",
-            code_register="",
+        PartijIdentificatorTypesValidator()(
+            partij_identificator={
+                "code_objecttype": PartijIdentificatorCodeObjectType.natuurlijk_persoon.value,
+                "code_soort_object_id": PartijIdentificatorCodeSoortObjectId.bsn.value,
+                "object_id": "296648875",
+                "code_register": "",
```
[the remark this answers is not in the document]
stevenbal: The only use case I could think of is to have `partijIdentificator: null` if `anderePartijIdentificator` is specified, but there doesn't seem to be any validation on either being filled, so I'm honestly not sure. The help text of `anderePartijIdentificator` is very vague as well (`Vrij tekstveld om de verwijzing naar een niet-voorgedefinieerd objecttype, soort objectID of Register vast te leggen.`), not sure how you would put all those things in a single field.

@danielmursa-dev can you create a separate issue to make these fields not accept empty values (code_objecttype, etc), to make `partijIdentificator` work when set to null (because I don't think that works now) and to add validation that either `partijIdentificator` or `anderePartijIdentificator` is specified

stevenbal (changes requested):

src/openklant/components/klantinteracties/tests/test_validators.py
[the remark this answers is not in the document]
stevenbal: So `object_id` would be left empty if `code_register=overig`? Could `partijIdentificator` being nullable be an oversight then? 
afbeelding

The more I look at this API spec, the more questions I start asking myself to be honest

PR #379, opened 2025-03-26
[#376] Fix query params with camelCase naming

### LINKED ISSUE #376: Query parameters for GET on partij-identificatoren don't seem to work?

2.6.1

In the [API-spec of version 2.6.1](https://redocly.github.io/redoc/?url=https://raw.githubusercontent.com/maykinmedia/open-klant/2.6.1/src/openklant/components/klantinteracties/openapi.yaml#tag/partij-identificatoren) there are listed several Query parameters on the partij-identificatoren endpoint. 

We have the impression that these don't work properly. 
In our 2.6.1-instance the GET call on https://openklant.dev.kiss-demo.nl/klantinteracties/api/v1/partij-identificatoren at this point returns 8 items
But the GET call on https://openklant.dev.kiss-demo.nl/klantinteracties/api/v1/partij-identificatoren?partijIdentificatorObjectId=69599068 also returns 8 items. 
And the GET call on https://openklant.dev.kiss-demo.nl/klantinteracties/api/v1/partij-identificatoren?partijIdentificatorCodeSoortObjectId=kvk_nummer also returns 8 items.

_No response_

_No response_

danielmursa-dev:
Fixes #376 

The bug was related to the camelCase naming of the parameters, in specific, if you execute the request with the camelCase, the request works fine, for example: 
`http://localhost:8000/klantinteracties/api/v1/partij-identificatoren?partij_identificator_code_register=hr`

Instead, it didn't not work with camelCase, because the `filter_backends` was updated with `DjangoFilterBackend` and not with the default `‘vng_api_common.filters.Backend’` which supports camelCase.

`http://localhost:8000/klantinteracties/api/v1/partij-identificatoren?partijIdentificatorCodeRegister=brp`

now both cases are allowed

### DIFF
--- a/src/openklant/components/klantinteracties/api/viewsets/klantcontacten.py
+++ b/src/openklant/components/klantinteracties/api/viewsets/klantcontacten.py
@@ -1,4 +1,3 @@
-from django_filters.rest_framework import DjangoFilterBackend
 from drf_spectacular.utils import extend_schema, extend_schema_view
 from rest_framework import mixins, viewsets
 from vng_api_common.pagination import DynamicPageSizePagination
@@ -178,7 +177,6 @@ class OnderwerpobjectViewSet(viewsets.ModelViewSet):
     serializer_class = OnderwerpobjectSerializer
     lookup_field = "uuid"
     pagination_class = DynamicPageSizePagination
-    filter_backends = [DjangoFilterBackend]
     filterset_fields = [
         "onderwerpobjectidentificator_object_id",
         "onderwerpobjectidentificator_code_objecttype",
@@ -221,7 +219,6 @@ class BijlageViewSet(viewsets.ModelViewSet):
     serializer_class = BijlageSerializer
     lookup_field = "uuid"
     pagination_class = DynamicPageSizePagination
-    filter_backends = [DjangoFilterBackend]
     filterset_fields = [
         "bijlageidentificator_object_id",
         "bijlageidentificator_code_objecttype",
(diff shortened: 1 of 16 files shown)


src/openklant/components/klantinteracties/api/viewsets/klantcontacten.py
```diff
@@ -221,7 +219,6 @@ class BijlageViewSet(viewsets.ModelViewSet):
     serializer_class = BijlageSerializer
     lookup_field = "uuid"
     pagination_class = DynamicPageSizePagination
-    filter_backends = [DjangoFilterBackend]
```
stevenbal, line 224: For my information, is there a default backend that is used now? It works but I'm not sure how :thinking:

stevenbal (approved):

src/openklant/components/klantinteracties/api/viewsets/klantcontacten.py
danielmursa-dev: Yes this one: https://github.com/maykinmedia/commonground-api-common/blob/f7814d22d6766f88f49688adb6af6d085b9f0458/vng_api_common/filters.py#L28

and here it is set https://github.com/maykinmedia/commonground-api-common/blob/f7814d22d6766f88f49688adb6af6d085b9f0458/vng_api_common/conf/api.py#L36

stevenbal: check, thanks for explaining

PR #384, opened 2025-04-03
:heavy_plus_sign: Add django-silk for performance profiling

stevenbal:
I noticed that some endpoints are missing prefetch_related/select_related, resulting in a lot of queries, django-silk can be used to spot this by setting the envvar `PROFILE=yes` and navigating to `/silk` after performing requests

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -16,8 +16,6 @@ asgiref==3.8.1
     #   django-cors-headers
 asn1crypto==1.5.1
     # via webauthn
-async-timeout==5.0.1
-    # via redis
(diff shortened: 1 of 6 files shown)


requirements/base.txt
```diff
@@ -16,8 +16,6 @@ asgiref==3.8.1
     #   django-cors-headers
 asn1crypto==1.5.1
     # via webauthn
-async-timeout==5.0.1
-    # via redis
```
stevenbal, line 20: Not sure why these were removed

danielmursa-dev (approved):

PR #397, opened 2025-04-14
:zap: [#378] Improve performance partij admin

### LINKED ISSUE #378: ValueError: 'Partij' instance needs to have a primary key value before this relationship can be used.

2.5.0 > 2.6

When a user adds a party under customer interactions and follows the steps below, they encounter a ValueError: 'Party' instance needs to have a primary key value before this relationship can be used.

[https://sentry.maykinmedia.nl/organizations/maykin-media/issues/403412/?project=430&query=is%3Aunresolved](url)

1. Click on Partij under Klantinteracties
2. Click on Partij toevoegen
3. Fill in the form in this way:
4. Image
5. Click on save

_No response_

danielmursa-dev:
Fixes #378 

**Changes**

- Improved validation of the admin form as it was done before (using reference to related objects).
- And added a new validation: because the data `voorkeurs_digitaal_adres` and `voorkeurs_rekeningnummer`, should be added after the creation of `partij (instance)`, otherwise the relationships don't exist.
- Simplified `partij.__str__()`

### DIFF
--- a/src/openklant/components/klantinteracties/admin/partijen.py
+++ b/src/openklant/components/klantinteracties/admin/partijen.py
@@ -1,6 +1,5 @@
 from django import forms
 from django.contrib import admin
-from django.core.exceptions import ValidationError
 from django.utils.translation import gettext_lazy as _
 
 from openklant.components.klantinteracties.models.rekeningnummers import Rekeningnummer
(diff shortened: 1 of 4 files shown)


stevenbal (changes requested):
@danielmursa-dev this does not address the performance of a GET on the change view for a Partij `/admin/klantinteracties/partij/<pk>/change/

In my local database I have about 1000 Partijen, and I get a lot of queries when I perform a GET
afbeelding

Could you look into what is causing these queries?

danielmursa-dev:
@stevenbal yes, confirm that the part that was causing many queries is the string representation of Partij, which makes many queries, so I removed the filters for `soort_partij` to improve performance.

Floris272 (approved):
~~The change views still does around 1800 queries because of the Onetoonefield inlines. But i have not found a quick fix, maybe the field can be moved to `Partij`? but that should probably be done in another ticket.~~

Its seems to be fixed, sorry my branch was not up to date.

stevenbal (approved):

PR #398, opened 2025-04-15
:memo: [#341] Fix help_text field in FilterSet

### LINKED ISSUE #341: No help texts / regex patterns seem to be picked up for these fields

2.5.0

It seems that for some fields that are not specified in the filterset, the help_text in the API specification is not taken automatically.
A solution may be here: https://django-filter.readthedocs.io/en/stable/guide/tips.html#adding-model-field-help-text-to-filters

Image

_No response_

_No response_

danielmursa-dev:
Fixes #341

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -57,7 +57,7 @@ click-plugins==1.1.1
     # via celery
 click-repl==0.3.0
     # via celery
-commonground-api-common==2.5.5
+commonground-api-common==2.6.2
     # via open-api-framework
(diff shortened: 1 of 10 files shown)


stevenbal (changes requested):
Looks good, could you also update the help text for `DigitaalAdres.adres`? This seems to still contain the regex before the changes made in https://github.com/maykinmedia/open-klant/issues/368

afbeelding

stevenbal (approved):

PR #412, opened 2025-05-08
Issue/337 incorrect helptext

### LINKED ISSUE #337: OnderwerpObject.onderwerpobjectidentificator help texts are incorrect?

2.5.0 / latest

I might be misunderstanding what an `OnderwerpObject` is, but from what I understand from the information model it is used to link KlantContacten to objects such as Zaken (https://vng-realisatie.github.io/klantinteracties/informatiemodel/semantisch_model)

In the help text of the identificator fields however, examples are given that refer to a Natuurlijk Persoon (by using BSN)
Image

Also since no enums exist for `codeRegister`, `codeObjecttype` and `codeSoortObjectId`, it is not clear what values those are capable of having

_No response_

I think there should be enums for the mentioned fields, similar to PartijIdentificator:

Image

I'm not sure what the values of those enums should be though, so this is up for discussion

TimdeBeer1:
Fixes #337 

**Changes**

Changed help text of the onderwerp object.


stevenbal (changes requested):
@TimdeBeer1 could you squash/rebase your commits into a single one? That makes the git history a bit cleaner :)

stevenbal:
@TimdeBeer1 I think something went wrong while trying to squash, generally what I do if I want to squash commits is:

Say I have commit history (from new to old)

- Commit A
- Commit B
- Commit C

And I want to create a single commit out of these three, I run `git reset --soft HEAD~2`, this removes the latest 2 commits, but keeps the changes locally, then you can `git add .` to add the changes and run `git commit --amend` to add these changes to commit C

You could also use `git rebase` (https://git-scm.com/docs/git-rebase), but that's usually a bit more complicated imo

stevenbal (approved):

PR #413, opened 2025-05-08
:speech_balloon: [#388] fix postal code validation message

### LINKED ISSUE #388: Postal code validation regex and error message are counter-intuitive
### Thema / Theme

Klantinteracties API

The current postal code regex requires a space between the numeric and alphabetic parts of the string:

```python
validate_postal_code = CustomRegexValidator(
    regex="^[1-9][0-9]{3} [A-Z]{2}$", message=_("Ongeldige postcode")
)
```
This leads to the counterintuitive result that `1234AB` is actually invalid, compounded by an error message that does not indicate how the string is malformed:

```
E                   openklant2.exceptions.BadRequest: status=400 code=400 title="Invalid input.":
E                   Invalid parameters:
E                   {'code': 'invalid',
E                    'name': 'correspondentieadres.postcode',
E                    'reason': 'Ongeldige postcode: 1234AB'}
```

My proposal would be to either allow both the space and space-less variety, or refining the validation to give a clearer error message (e.g. rather than `Ongeldige postcode`, use `Postcode moet formaat 1234AB gebruiken`).

_No response_

_No response_

TimdeBeer1:
Fixes #388

**Changes**

Changed postal code validation message so it shows a space is needed.


src/openklant/utils/validators.py
```diff
@@ -67,7 +67,8 @@ def __call__(self, value):
 
 
 validate_postal_code = CustomRegexValidator(
-    regex="^[1-9][0-9]{3} [A-Z]{2}$", message=_("Ongeldige postcode")
+    regex="^[1-9][0-9]{3} [A-Z]{2}$",
+    message=_("Postcode moet aan het volgende formaat voldoen: 1234 AB"),
```
stevenbal, line 71: Let's make this as explicit as possible
```suggestion
    message=_("Postcode moet aan het volgende formaat voldoen: `1234 AB` (met spatie)"),
```

stevenbal (changes requested):

src/openklant/components/contactgegevens/migrations/0008_alter_organisatie_adres_postcode_and_more.py
```diff
@@ -0,0 +1,46 @@
+# Generated by Django 4.2.20 on 2025-05-09 13:21
+
+from django.db import migrations, models
+import openklant.utils.validators
+
+
+class Migration(migrations.Migration):
```
stevenbal, line 7: @TimdeBeer1 generally it's better to have fewer migration files if it's possible, to keep the migration history clean (and also make migrations faster). Especially in this case, because the second migration is a fix based on PR feedback.

You can change this into a single migration by first undoing the migrations locally:

```
python src/manage.py migrate contactgegevens 0006_remove_organisatie_land_remove_persoon_land_and_more
```

Then you can delete the two new migration files `0007` and `0008` and re-run `src/manage.py makemigrations` to generate a single migration file. The same can be done for the klantinteracties app.

And ideally, since the change from the feedback to the helptext was quite small, I'd amend the original commit with the changes, to keep the commit history clean as well

stevenbal (changes requested):

PR #420, opened 2025-05-16
:fire: [#351] Remove overig from partijidentificator

### LINKED ISSUE #351: Remove possibility to record "overige" PartijIdentificatoren
In the current API spec it is not effectively possible to record "overige" (i.e. non-formal, such as bsn or kvk) identificatoren. This is for two reasons:

1. The various enums for register, object type, etc. contain an 'overig' option, but this is incomplete: you have no good place to specify what _kind_ of overig object it is (except possibly in the `anderePartijIdentificator` field, but that does not seem very consistent)
2. Now that the various meta-attributes are unique together, you can also only add a single "overig" identifier (for a given identifier value).

This means that, currently, "overige" identificatoren are broken. We need a better solution for this, but we also do not want to delay the release of 2.6.0 for this. The proposal here is therefore to:

1. Remove 'overig' from all the identificator enums
2. Remove the field `anderePartijIdentificator`
3. Make `partijIdentificator` [non-nullable in the serializer ](https://github.com/maykinmedia/open-klant/blob/1e01be8c9d5ef9291662d1ad308647b2240854e3/src/openklant/components/klantinteracties/api/serializers/partijen.py#L388)

So that the current object is at least coherent and consistent for the "formal" registrations. We will then re-introduce a mechanism for overige identificatoren after we agree on a proposal with the various stakeholders. This is preferable to a release in which overige identificatoren are effectively broken.

Current API spec:
Image

danielmursa-dev:
Fixes #351 

- Remove `overig` from partijidentificator 
- Set `andere_partij_identificator` as deprecated
- Update tests


src/openklant/components/klantinteracties/api/schema.py
```diff
@@ -185,10 +185,9 @@
 
 | codeRegister  | codeObjecttype       | codeSoortObjectId                |
 |---------------|--------------------------|------------------------------|
-| brp           | natuurlijk_persoon       | bsn, overig                  |
-| hr            | niet_natuurlijk_persoon  | rsin, kvk_nummer, overig     |
-| hr            | vestiging                | vestigingsnummer, overig     |
-| overig        | –                        | –                            |
+| brp           | natuurlijk_persoon       | bsn,                         |
+| hr            | niet_natuurlijk_persoon  | rsin, kvk_nummer,            |
+| hr            | vestiging                | vestigingsnummer,            |
```
stevenbal, line 190:
```suggestion
| hr            | niet_natuurlijk_persoon  | rsin, kvk_nummer             |
| hr            | vestiging                | vestigingsnummer             |
```

src/openklant/components/klantinteracties/api/serializers/partijen.py
```diff
@@ -387,6 +387,7 @@ class Meta:
         validators = []
 
 
+@extend_schema_serializer(deprecate_fields=["andere_partij_identificator"])
```
stevenbal, line 390: Could you also mark the query parameter as deprecated?

afbeelding

CHANGELOG.rst
```diff
@@ -2,6 +2,15 @@
...
 ==============
 
+2.x.x
+=====
+
+* [:open-klant:`351`] Remove overig from partijidentificator
+
+.. warning::
+
+    The field ``andere_partij_identificator`` for ``/partijidentificator`` endpoint is now deprecated and will be removed in the next release
```
stevenbal, line 12:
```suggestion
    The field ``anderePartijIdentificator`` for ``/partij-identificatoren`` endpoint is now deprecated and will be removed in the next major release
```

CHANGELOG.rst
```diff
@@ -2,6 +2,15 @@
 Change history
 ==============
 
+2.x.x
+=====
+
+* [:open-klant:`351`] Remove overig from partijidentificator
```
stevenbal, line 8:
```suggestion
* [:open-klant:`351`] Remove overig from partijidentificator and deprecate `anderePartijIdentificator`
```

Floris272 (approved):

src/openklant/components/klantinteracties/api/viewsets/partijen.py
```diff
@@ -221,7 +221,10 @@ class CategorieViewSet(viewsets.ModelViewSet):
     permission_classes = (TokenPermissions,)
 
 
-@extend_schema(tags=["partij-identificatoren"])
+@extend_schema(
+    tags=["partij-identificatoren"],
+    parameters=[OpenApiParameter(name="anderePartijIdentificator", deprecated=True)],
```
stevenbal, line 226: This is now added for each operation, also create for example, even though it only is relevant for the list operation

afbeelding

If you add `parameters=[]` to the `list=extend_schema(...)` below, it will only be added for the list operation

stevenbal (changes requested):
Small thing remaining, looks good otherwise

src/openklant/components/klantinteracties/api/viewsets/partijen.py
danielmursa-dev: Thanks :+1:

stevenbal (approved):

PR #423, opened 2025-05-16
:art: [maykinmedia/open-api-framework#133] Refactor code with ruff

danielmursa-dev:
Fixes #[maykinmedia/open-api-framework#133]

Workflow: https://github.com/maykinmedia/open-api-workflows/pull/23

- [x] Install `ruff` and uninstall `black`, `isort`, `flake8`
- [x] Update `code-quality` workflow version
- [x] Format code after ruff checks
- [x] Update README, index.rst
- [x] Set `pre-commit` and update docs
- [x] Add `.git-blame-ignore-revs`


src/openklant/accounts/management/commands/createinitialsuperuser.py
```diff
@@ -33,7 +33,7 @@ def handle(self, *args, **options):
         User.objects.create_superuser(username=username, email=email, password=password)
 
         try:
-            link = f'{settings.ALLOWED_HOSTS[0]}{reverse("admin:index")}'
+            link = f"{settings.ALLOWED_HOSTS[0]}{reverse('admin:index')}"
```
stevenbal, line 36: Could you add a file `.git-blame-ignore-revs` in the root of the repo and add the full commit hash of the commit that applies the formatting changes? That way git blame will exclude it

src/openklant/utils/validators.py
```diff
@@ -76,7 +76,8 @@ def __call__(self, value):
 )
 
 validate_no_space = CustomRegexValidator(
-    regex="^[\S]+$", message=_("Geen spaties toegestaan")  # noqa
+    regex="^[\S]+$",
+    message=_("Geen spaties toegestaan"),  # noqa
```
stevenbal, line 80: I'm not sure if the `noqa` is needed anymore, maybe we can remove it?

src/openklant/migration/client.py
```diff
@@ -35,10 +39,8 @@ def _request(
         except requests.RequestException:
             logger.exception(f"{method} request failed for {url}")
 
-            try:
+            with contextlib.suppress(requests.RequestException, UnboundLocalError):
```
stevenbal, line 42: I don't really understand why this was the way it was previously to be honest. I think the logger.debug statement should just be in an `else` statement after the except, because that is only evaluated if no exceptions occurred (https://docs.python.org/3/reference/compound_stmts.html#else-clause)

```
try:
    # make request
except requests.RequestException:
    logger.exception()
else:
    logger.debug()
```

stevenbal (changes requested):

PR #436, opened 2025-05-30
:lipstick: Fix light/dark themes for admin

TimdeBeer1:
Fixes #https://github.com/maykinmedia/open-api-framework/issues/149

**Changes**

 Fixed light/dark theme in admin UI with the missing toggle.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,7 +50,11 @@ jobs:
 
     strategy:
       matrix:
-        postgres: ['13', '14', '15', '16', '17']
+        postgres: ["14", "15", "16", "17"]
+        use_pooling: [false]
+        include:
+          - postgres: "17"
+            use_pooling: true
 
     services:
       postgres:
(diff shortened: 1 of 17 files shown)


Floris272:
@TimdeBeer1 
With browser set to dark mode and OK to light mode:
image

(other projects do not seem to have this issue)

Maybe its the merge conflict.

Floris272:
image

image

- Open klant has `_vars.scss` located directly in `scss`. `_admin_theme.scss` for examples links to  it `@import '../vars';
` so i think `scss/admin/_vars.scss` is currently unused.

- Open klant also seems to already have some dark mode stuff, `@media (prefers-color-scheme..` is used in _app_overrides.scss for example. The other projects don't seem to have this. Maybe this causes the issue. 

- I quickly tested it in main and i had the same issues so maybe the old stuff can just be removed but please double check.

stevenbal (approved):

PR #449, opened 2025-06-26
:whale: Add celery container to docker-compose.yml

stevenbal:
to be a more realistic setup, because Open Klant uses celery to send notifications


stevenbal:
Noticed this was missing while testing structlog

PR #456, opened 2025-07-14
:recycle: [open-api-framework#67] add optional dependencies

TimdeBeer1:
Fixes maykinmedia/open-api-framework#67

**Changes**

added the optional dependencies to oaf in base.in


stevenbal (changes requested):
@TimdeBeer1 I released version 0.12.0 of OAF (which also adds the connection pooling envvars), could you use that version and remove the DB pooling envvars (including DB_CONN_MAX_AGE) from the Open Klant settings?

PR #464, opened 2025-07-28
:sparkles: [#437] Add verificatie_datum to DigitaalAdres

### LINKED ISSUE #437: Digitaal adres - Verificatie datum
### Thema / Theme

Klantinteracties API

Use case:
In our gemeente we already have registered 700.000 mail addresses of our clients which did not require verification at the time.
When going to OpenKlant 2 we would like to migrate these users. The thing is we need to discern these addresses between verified and non verified.
We propose to create a new field 'verification date' (which might lead to an implicit 'verified' boolean on the actual registration) so that we can see if / when a user has verified their address.

For future requirements we prefer a date instead of a boolean so we could verify the address after a year for example.

_No response_

_No response_

OlhaZahoruiko:
Fixes #437

**Changes**

[Describe the changes here]


Floris272 (changes requested):

danielmursa-dev (approved):

src/openklant/components/klantinteracties/api/filterset/digitaal_adres.py
```diff
@@ -131,6 +131,7 @@ class Meta:
             "is_standaard_adres",
             "omschrijving",
             "referentie",
+            "verificatie_datum",
```
stevenbal, line 134: This means we can only filter on exact verification dates, which I don't think is that useful. I think we should at least add `gt`, `lt`, `gte` and `lte` to make it a bit more useful.

I also think it would be good to have a read only field `geverifieerd` + a filter for that

src/openklant/components/klantinteracties/migrations/0037_digitaaladres_verificatie_datum.py
```diff
@@ -0,0 +1,24 @@
...
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("klantinteracties", "0036_alter_partij_voorkeurs_digitaal_adres_and_more"),
+    ]
+
+    operations = [
```
stevenbal, line 12: Alex also mentioned in the issue that it should be possible to verify DigitaalAdressen in bulk (because by default existing adressen won't be verified), so I think it would be good to have a management command to do this

stevenbal: added command `bulk_verify_digitaal_adressen`

stevenbal (approved):

Floris272 (commented):
How would the command be used? 

I suppose if all addresses that are known to be verified are migrated first, you could run the command. But after a migrations that contains addresses that could be unverified it can not be used again?

Issue #470, opened 2025-08-06
Validation error for country code should warn for upper/lower case issues


swrichards:
### Thema / Theme

Klantinteracties API

### Omschrijving / Description

Currently the country code is expected to be uppercase, and if a correct country code is provided in lowercase, it is rejected as an invalid code: https://github.com/maykinmedia/open-klant/blob/adf4d6cad99ee43758fb684883db449b18de4660/src/openklant/utils/validators.py#L16-L19

It would be useful for the user to offer a suggestion if a country code is correct except for the case.

PR #485, opened 2025-08-25
:bookmark: Release 2.12.0

danielmursa-dev:
Fixes #

**Changes**

[Describe the changes here]

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,27 @@
 Change history
 ==============
 
+2.12.0
+======
+*August 25, 2025*
+
+**New features**
+
+* [:open-klant:`419`] Added query parameter on ``onderwerpobjecten`` for searching klantcontacten by referred klantcontact UUID.
+
+**Documentation**
+
+* [:open-api-framework:`159`] Add functionality to create model image (see :ref:`uml_diagrams`)
+
+**Maintenance**
+
+* Upgrade python dependencies
+
+  * ``zgw-consumers`` to 1.0.0
+  * ``commonground-api-common`` to 2.9.0
+
+* [:open-api-framework:`179`] Add monkeypatch to requests applying a default timeout to all requests calls
+
 2.11.1
 ======
 *August 8, 2025*
--- a/src/openklant/conf/api.py
+++ b/src/openklant/conf/api.py
@@ -4,7 +4,7 @@ from vng_api_common.conf.api import *  # noqa - imports white-listed
 # Remove the reference - we don't have a single API version.
 del API_VERSION  # noqa
 
-KLANTINTERACTIES_API_VERSION = "0.3.0"
+KLANTINTERACTIES_API_VERSION = "0.3.1"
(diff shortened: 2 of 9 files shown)


Floris272 (changes requested):
The api spec does not load for me because the styling is blocked by csp, Got it working using `CSP_INCLUDE_NONCE_IN.remove("style-src")` but there is probably a better solution

CHANGELOG.rst
```diff
@@ -2,6 +2,28 @@
 Change history
 ==============
 
+2.12.0
+======
+*August 26, 2025*
+
+**New features**
+
+* [:open-klant:`419`] Added query parameter on ``onderwerpobjecten`` for searching klantcontacten by referred klantcontact UUID.
```
stevenbal, line 11: could you also add the exact name of the query param here?

CHANGELOG.rst
```diff
@@ -2,6 +2,28 @@
...
+
```
stevenbal, line 12: I think this is missing #470 under bugfixes/qol

src/openklant/conf/api.py
```diff
@@ -4,7 +4,7 @@
 # Remove the reference - we don't have a single API version.
 del API_VERSION  # noqa
 
-KLANTINTERACTIES_API_VERSION = "0.3.0"
+KLANTINTERACTIES_API_VERSION = "0.3.1"
```
stevenbal, line 7: we added a new query param, so let's make it a minor version bump
```suggestion
KLANTINTERACTIES_API_VERSION = "0.4.0"
```

stevenbal (changes requested):

stevenbal:
@Floris272 I think I've seen the same error before locally, it doesn't seem to happen when running with docker compose, not sure why

danielmursa-dev:
image

@Floris272 @stevenbal , yes confirm, I also saw this error before, in locally, there is `DEBUG=True`, Django adds a `nonce-{token}` to the `CSP`, which causes `unsafe-inline` to be ignored, i don't know why? instead on `docker` the debugging is `False`

Floris272 (approved):

CHANGELOG.rst
```diff
@@ -2,6 +2,37 @@
...
+* [:open-klant:`419`] Added query parameters on `onderwerpobjecten` for searching klantcontacten by referred klantcontact:
+
+    * ``klantcontact__uuid``
+    * ``klantcontact__url``
+    * ``was_klantcontact__uuid``
+    * ``was_klantcontact__url``
+
+**Bugfixes/QOL**
+
+* [:open-klant:`470`] Chang ``countrycode`` validation message to warn when the value has incorrect casing.
```
stevenbal, line 20:
```suggestion
* [:open-klant:`470`] Change ``countrycode`` validation message to warn when the value has incorrect casing.
```

stevenbal (approved):
@danielmursa-dev could you make an issue for the CSP bug in open-api-framework?

PR #490, opened 2025-09-03
:bug: [#486] Ensure default_admin_index.json is loaded via signal

### LINKED ISSUE #486: Ensure default_admin_index.json is loaded via signal as OF/OZ

2.11.0

Issue discovered https://github.com/maykinmedia/open-klant/pull/482#discussion_r2298289918 while upgrading `mozilla-django-oidc-db` to 0.25.1
The admin JSON file (`default_admin_index`) was loaded only during migration `0002_auto_20211116_0918` and not on every migration like in OF/OZ, which should be handled via a signal.

_No response_

_No response_

OlhaZahoruiko:
Fixes #486

**Changes**

[Describe the changes here]

### DIFF
--- a/src/openklant/accounts/signals.py
+++ b/src/openklant/accounts/signals.py
@@ -1,0 +1,19 @@
+from io import StringIO
+
+from django.apps import apps
+from django.contrib.contenttypes.management import create_contenttypes
+from django.core.management import call_command
+
+from django_admin_index.models import AppGroup
+
+
+def update_admin_index(sender, **kwargs):
+    AppGroup.objects.all().delete()
+
+    project_name = __name__.split(".")[0]
+
+    for app_config in apps.get_app_configs():
+        if app_config.name.startswith(project_name):
+            create_contenttypes(app_config, verbosity=0)
+
+    call_command("loaddata", "default_admin_index", verbosity=0, stdout=StringIO())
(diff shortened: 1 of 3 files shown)


danielmursa-dev (changes requested):
I checked and it seems that it doesn't work because, when you run 
`post_migrate.connect(update_admin_index, sender=self)`, this `post_migrate` is only executed if there is a migration for this app, so there are no migrations in this `utils` package.
Solutions:
- Or set the `sender=None` -> `post_migrate.connect(update_admin_index, sender=None)`, so this signal starts after all apps, but I don't think this is a good solution because this command would be executed `n` times.
- or move this in `accounts` app as in open-forms here: https://github.com/open-formulieren/open-forms/blob/3a24ebd2735dc79ac7769104d8d855ba149aa347/src/openforms/accounts/apps.py and should works now only one time

src/openklant/accounts/signals.py
```diff
@@ -0,0 +1,21 @@
+# SPDX-License-Identifier: EUPL-1.2
+# Copyright (C) 2022 Dimpact
```
danielmursa-dev, line 2: Not necessary in OpenKlant
```suggestion
```

stevenbal (approved):

PR #502, opened 2025-09-16
:sparkles: [#501] migrate_to_v2 migrate phone numbers

### LINKED ISSUE #501: `migrate_to_v2` script should also migrate phone numbers
### Thema / Theme

Klantinteracties API

In the current version of the `migrate_to_v2` command, phone numbers are not migrated. We should add a second script to make sure this is done as well

TODO:
- [ ] Rename `Klant.to_digitaal_adres` to `Klant.to_digitaal_adres_email`
- [ ] Add `Klant.to_digitaal_adres_phonenumber` and the required attributes to construct the DigitaalAdres for the phone number
- [ ] Add a second management command `migrate_to_v2_phonenumbers` that does the following:
  - [ ] Should have the same arguments and logic to generate API tokens as the original script
  - [ ] Should also retrieve all the klanten from the v1 instance, but instead of creating Partijen and DigitaalAdressen with emails, it should fetch the existing partij (because it was already created by running `migrate_to_v2` earlier), it should check if there are **no** existing DigitaalAdressen with `soortDigitaalAdres=telefoonnummer` and `referentie=portaalvoorkeur` and if that is the case it should use `Klant.to_digitaal_adres_phonenumber` and create a DigitaalAdres with the phone number and link it to the existing partij (by setting `verstrekt_door_partij`)
  - [ ] `DigitaalAdres.referentie` must also be set to `portaalvoorkeur` (same as done previously for emails)
- [ ] Add testcases:
  - [ ] Klant with telefoonnummer leads to a new DigitaalAdres
  - [ ] Klant without telefoonnumer is skipped
  - [ ] Klant with telefoonnummer but with an existing DigitaalAdres with  `soortDigitaalAdres=telefoonnummer` and `referentie=portaalvoorkeur` is skipped

_No response_

_No response_

OlhaZahoruiko:
Fixes #501

**Changes**

[Describe the changes here]

### DIFF
--- a/migration/cassettes/test_telefoonnummer.yaml
+++ b/migration/cassettes/test_telefoonnummer.yaml
@@ -1,0 +1,88 @@
+interactions:
+- request:
+    body: null
+    headers:
+      Accept:
+      - '*/*'
+      Accept-Encoding:
+      - gzip, deflate
+      Connection:
+      - keep-alive
+      User-Agent:
+      - python-requests/2.32.3
+    method: GET
+    uri: http://localhost:8000/klanten/api/v1/klanten
+  response:
+    body:
+      string: '{"count":1,"next":null,"previous":null,"results":[{"url":"http://localhost:8000/klanten/api/v1/klanten/dbfdb3f2-e153-49ea-9e26-e55220605c14","bronorganisatie":"215906846","klantnummer":"0","bedrijfsnaam":"","functie":"","websiteUrl":"","voornaam":"","voorvoegselAchternaam":"","achternaam":"","telefoonnummer":"0612345678","emailadres":"example@maykinmedia.nl","adres":null,"subject":"","subjectType":"natuurlijk_persoon","aanmaakkanaal":"","geverifieerd":false,"subjectIdentificatie":{"inpBsn":"024325818","anpIdentificatie":"0","inpANummer":"","geslachtsnaam":"Greene","voorvoegselGeslachtsnaam":"","voorletters":"","voornamen":"Lori","geslachtsaanduiding":"","geboortedatum":"1981-04-15","verblijfsadres":null,"subVerblijfBuitenland":null}}]}'
+    headers:
+      API-version:
+      - 1.0.0
+      Allow:
+      - GET, POST, HEAD, OPTIONS
+      Content-Length:
+      - '726'
+      Content-Type:
+      - application/json
+      Referrer-Policy:
+      - same-origin
+      X-Content-Type-Options:
+      - nosniff
+      X-Frame-Options:
+      - DENY
+      X-XSS-Protection:
+      - 1; mode=block
+    status:
+      code: 200
+      message: OK
+- request:
+    body: null
+    headers:
+      Accept:
+      - '*/*'
+      Accept-Encoding:
+      - gzip, deflate
+      Authorization:
+      - Token e2a63da7e0f37f5463ba0493744a937fc1adac02
+      Connection:
+      - keep-alive
+      User-Agent:
+      - python-requests/2.32.4
+    method: GET
+    uri: http://localhost:8005/klantinteracties/api/v1/partijen?nummer=&expand=digitaleAdressen
+  response:
+    body:
+      string: '{"count":1,"next":null,"previous":null,"results":[{"uuid":"44a7b364-e464-449a-a106-ffddcef06f67","url":"http://localhost:8005/klantinteracties/api/v1/partijen/44a7b364-e464-449a-a106-ffddcef06f67","nummer":"12345","interneNotitie":"","betrokkenen":[],"categorieRelaties":[],"digitaleAdressen":[],"voorkeursDigitaalAdres":{"uuid":"631fcdaf-702a-46d3-8639-f1b458fac49c","url":"http://localhost:8005/klantinteracties/api/v1/digitaleadressen/631fcdaf-702a-46d3-8639-f1b458fac49c"},"vertegenwoordigden":[],"rekeningnummers":[],"voorkeursRekeningnummer":null,"partijIdentificatoren":[],"soortPartij":"persoon","indicatieGeheimhouding":true,"voorkeurstaal":"","indicatieActief":false,"bezoekadres":{"nummeraanduidingId":"","straatnaam":"","huisnummer":null,"huisnummertoevoeging":"","postcode":"","stad":"","adresregel1":"","adresregel2":"","adresregel3":"","land":""},"correspondentieadres":{"nummeraanduidingId":"","straatnaam":"","huisnummer":null,"huisnummertoevoeging":"","postcode":"","stad":"","adresregel1":"","adresregel2":"","adresregel3":"","land":""},"partijIdentificatie":null,"_expand":{"digitaleAdressen":[]}}]}'
+    headers:
+      API-version:
+      - 0.4.0
+      Allow:
+      - GET, POST, HEAD, OPTIONS
+      Content-Length:
+      - '1119'
+      Content-Security-Policy:
+      - 'worker-src ''self'' blob:; style-src ''self'' ''unsafe-inline'' fonts.googleapis.com;
+        default-src ''self''; script-src ''self'' ''unsafe-inline''; img-src ''self''
+        data: cdn.redoc.ly; frame-ancestors ''none''; object-src ''none''; frame-src
+        ''self''; form-action ''self''; font-src ''self'' fonts.gstatic.com; base-uri
+        ''self'''
+      Content-Type:
+      - application/json
+      Cross-Origin-Opener-Policy:
+      - same-origin
+      Date:
+      - Wed, 17 Sep 2025 08:19:32 GMT
+      Referrer-Policy:
+      - same-origin
+      Server:
+      - WSGIServer/0.2 CPython/3.12.3
+      Vary:
+      - origin
+      X-Content-Type-Options:
+      - nosniff
+      X-Frame-Options:
+      - DENY
+    status:
+      code: 200
+      message: OK
+version: 1
--- a/src/openklant/management/commands/migrate_to_v2_phonenumbers.py
+++ b/src/openklant/management/commands/migrate_to_v2_phonenumbers.py
@@ -3,7 +3,7 @@
 
 
 import os
-from dataclasses import asdict, fields as dataclass_fields
+from dataclasses import fields as dataclass_fields
 from typing import Any, Tuple
 from urllib.parse import parse_qs, urlparse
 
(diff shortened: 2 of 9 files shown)


src/openklant/management/commands/migrate_to_v2_phonenumbers.py
```diff
@@ -0,0 +1,226 @@
...
+    updated_klanten = []
+
+    for klant in klanten:
+        telefoonnummer = klant.to_digitaal_adres_phonenumber()
+
+        if not telefoonnummer:
+            continue
+
+        existing_partijen = openklant_client.retrieve(
+            PARTIJEN_PATH, params={"nummer": klant._get_subject().nummer}
```
stevenbal, line 106: In order to check the information for the related DigitaleAdressen, you'll have to add `expand=digitaleAdressen` to the params as well

src/openklant/tests/test_migrate.py
```diff
@@ -689,3 +693,74 @@ def test_rsin(self):
...
+        subject = Partij(nummer="12345", soort_partij=SoortPartij.persoon)
+        klant_with_phone._get_subject = lambda: subject
+
+        with patch(
+            "openklant.management.commands.migrate_to_v2_phonenumbers.OpenKlantClient"
+        ) as MockClient:
+            mock_client = MockClient.return_value
+            mock_client.retrieve.return_value = [
+                {"uuid": str(subject.uuid), "digitale_adressen": []}
+            ]
```
stevenbal, line 710: It's a bit unfortunate that for the `migrate_to_v2` tests we didn't really have a proper docker-compose setup with Open Klant v1, but I'd rather use VCR here than mocking stuff like this. Making this into a proper setup with VCR with a docker-compose will be annoying, because it's difficult to unittest different cases since we make calls to the list endpoint. Also this script is only a one time thing for Den Haag, so instead you could copy one of the existing cassettes and modify it to add a phonenumber

Also instead of calling `_save_klanten_phonenumbers`, could you make sure the tests are run the same as the `migrate_to_v2` tests above (by using `call_command`)?

src/openklant/tests/test_migrate.py
```diff
@@ -689,3 +693,74 @@ def test_rsin(self):
         )
 
         self.assertEqual(output, [f"{self.live_server_url}{partij_url}"])
+
+    def test_klant_with_telefoonnummer_creates_digitaaladres(self):
+        # Klant with telefoonnummer creates a new DigitaalAdres
+        klant_with_phone = Klant(telefoonnummer="0612345678")
+
+        subject = Partij(nummer="12345", soort_partij=SoortPartij.persoon)
```
stevenbal, line 701: Instead of instantiating these directly, you'll probably want to use factories. I don't think you'll have to instantiate the klant, because the klant data is retrieved with a request

src/openklant/tests/test_migrate.py
```diff
@@ -689,3 +693,74 @@ def test_rsin(self):
...
+            "referentie": "portaalvoorkeur",
+        }
+
+        with patch(
+            "openklant.management.commands.migrate_to_v2_phonenumbers.OpenKlantClient"
+        ) as MockClient:
+            mock_client = MockClient.return_value
+            mock_client.retrieve.return_value = [
+                {"uuid": str(subject.uuid), "digitale_adressen": [existing_address]}
+            ]
```
stevenbal, line 760: Instead of mocking the v2 responses, you could use a `DigitaalAdresFactory` to create the telefoonnummer with portaalvoorkeur as referentie, and just let the call to the API be done like it would be done in other API unittests

stevenbal (changes requested):

migration/cassettes/test_telefoonnummer.yaml
```diff
@@ -0,0 +1,88 @@
...
+      Accept-Encoding:
+      - gzip, deflate
+      Authorization:
+      - Token e2a63da7e0f37f5463ba0493744a937fc1adac02
+      Connection:
+      - keep-alive
+      User-Agent:
+      - python-requests/2.32.4
+    method: GET
+    uri: http://localhost:8005/klantinteracties/api/v1/partijen?nummer=&expand=digitaleAdressen
```
stevenbal, line 52: I see these query params in the URI here, but in the code I don't see `expand` being used. To make sure this works correctly, I'd say it's best to locally spin Open Klant v1, delete the cassettes and go over all of the three tests individually.

You could create a Klant with the required values (e.g. with telefoonnummer) and then run that specific test, then you can move to the next and modify the Klant accordingly. It's not ideal, but then at least it's guaranteed that it works correctly.

Docker compose you could use:
```
# Stripped version of https://github.com/maykinmedia/open-klant/blob/1.0.0/docker-compose.yml

services:
  db:
    image: postgres
    environment:
      - POSTGRES_HOST_AUTH_METHOD=trust
    volumes:
      - ./docker-init-db.sql:/docker-entrypoint-initdb.d/init_db.sql:ro
    # Note the missing database data directory. This make it easier to rerecord
    # cassettes as it will start with a clean database whenever
    # `docker compose down` is used (e.g the container is recreated on the next
    # startup).

  web:
    image: maykinmedia/open-klant:1.0.0
    environment:
      - DJANGO_SETTINGS_MODULE=openklant.conf.docker
      - IS_HTTPS=no
      - DB_NAME=postgres
      - DB_USER=postgres
      - DB_HOST=db
      - ALLOWED_HOSTS=*
      - CACHE_DEFAULT=redis:6379/0
      - CACHE_AXES=redis:6379/0
      - SECRET_KEY=${SECRET_KEY:-django-insecure-$8s@b*ds4t84-q_2#c0j0506@!l2q6r5_pq5e!vm^_9c*#^66b}
      - CELERY_BROKER_URL=redis://redis:6379/0
      - CELERY_RESULT_BACKEND=redis://redis:6379/0
      - NOTIFICATIONS_DISABLED=True
    volumes:
      - ./fixtures:/app/src/openklant/fixtures/migration:ro
    ports:
      - 8000:8000
    depends_on:
      - db
      - redis

  redis:
    image: redis

volumes:
  log:
```

src/openklant/management/commands/migrate_to_v2_phonenumbers.py
```diff
@@ -0,0 +1,244 @@
...
+
+
+def _retrieve_klanten(url: str, access_token: str) -> Tuple[list[Klant], str | None]:
+    klanten_path = "/klanten/api/v1/klanten"
+
+    _url = urlparse(url)
+    _params = parse_qs(_url.query)
+
+    client = LegacyOpenKlantClient(f"{_url.scheme}://{_url.netloc}", access_token)
+    response_data = client.retrieve(_url.path or klanten_path, params=_params)
```
stevenbal, line 55: Here you'll also have to add the expand on `digitaleAdressen`

src/openklant/management/commands/migrate_to_v2_phonenumbers.py
```diff
@@ -0,0 +1,244 @@
...
+            referentie="portaalvoorkeur",
+        ).first()
+
+        if existing_da:
+            logger.info(
+                "digitaaladres_exists",
+                partij=partij.pk,
+                nummer=nummer,
+            )
+            continue
```
stevenbal, line 139: Currently this does up to two queries per `klant` in `klanten`, which can become quite slow if there is a decent number of klanten.

You could simplify this by doing a single query on `Partij` to figure out which `Partijen` already have a telefoonnummer marked as portaalvoorkeur (something like `Partij.objects.filter(digitaaladres__referentie="portaalvoorkeur", digitaaladres__soort_digitaal_adres=SoortDigitaalAdres.telefoonnumer)`) and then you could skip those partijen in the loop

src/openklant/management/commands/migrate_to_v2_phonenumbers.py
```diff
@@ -0,0 +1,244 @@
...
+            application=MIGRATION_TOKEN_IDENTIFIER
+        )
+
+        token_auth = dummy_tokens.first()
+        deletion_tokens = dummy_tokens.exclude(pk=token_auth.pk)
+        deleted_token_ids = list(deletion_tokens.values_list("id", flat=True))
+        deletion_tokens.delete()
+
+        logger.warning("removed_migration_dummy_tokens", tokens=deleted_token_ids)
+    return token_auth.token
```
stevenbal, line 181: Is this the same as `_generate_dummy_token` in `migrate_to_v2`? If so, could you move them to the `utils.py` file in `openklant/migration`?

src/openklant/migration/v1/data.py
```diff
@@ -177,11 +179,22 @@ def set_from_external_subject(self, subject_data: dict) -> None:
...
+            adres=self.emailadres, soort_digitaal_adres=SoortDigitaalAdres.email
+        )
+
+    def to_digitaal_adres_phonenumber(self) -> DigitaalAdres | None:
+        if not self.telefoonnummer:
+            return
+        return DigitaalAdres(
+            adres=self.telefoonnummer,
+            soort_digitaal_adres=SoortDigitaalAdres.telefoonnummer,
+            omschrijving="Telefoonnummer",
```
stevenbal, line 196: Since `soort_digitaal_adres` already contains the same, let's not include this
```suggestion
```

stevenbal (changes requested):

src/openklant/management/commands/migrate_to_v2_phonenumbers.py
```diff
@@ -0,0 +1,219 @@
...
+
+def _retrieve_klanten(url: str, access_token: str) -> Tuple[list[Klant], str | None]:
+    klanten_path = "/klanten/api/v1/klanten"
+
+    _url = urlparse(url)
+    _params = parse_qs(_url.query)
+
+    client = LegacyOpenKlantClient(f"{_url.scheme}://{_url.netloc}", access_token)
+    response_data = client.retrieve(
+        _url.path or klanten_path, params={**_params, "expand": "digitaleAdressen"}
```
stevenbal, line 56: Sorry, I was wrong about this, I forgot that this is the API call to v1 and `digitaleAddressen` don't exist there, so this can be removed :facepalm: 

```suggestion
        _url.path or klanten_path, params=_params
```

src/openklant/migration/v1/data.py
```diff
@@ -177,11 +179,21 @@ def set_from_external_subject(self, subject_data: dict) -> None:
...
 
-        return DigitaalAdres(adres=self.emailadres)
+        return DigitaalAdres(
+            adres=self.emailadres, soort_digitaal_adres=SoortDigitaalAdres.email
+        )
+
+    def to_digitaal_adres_phonenumber(self) -> DigitaalAdres | None:
+        if not self.telefoonnummer:
+            return
+        return DigitaalAdres(
```
stevenbal, line 193: I see in the dataclass the default for omschrijving is `Emailadres`, could you remove that default?

image

src/openklant/tests/test_migrate.py
```diff
@@ -689,3 +693,78 @@ def test_rsin(self):
         )
 
         self.assertEqual(output, [f"{self.live_server_url}{partij_url}"])
+
+    def test_telefoonnummer(self):
```
stevenbal, line 697: Just to be sure, could you also add this test for a Klant with type `niet_natuurlijk_persoon`?

stevenbal (changes requested):
Small stuff remaining

stevenbal (approved):

PR #511, opened 2025-09-25
:bug: [#510] Add missing raw_id_fields for PartijAdmin

### LINKED ISSUE #510: PartijAdmin detail page is very slow if the database has a decent number of Partijen

2.12.1

Taiga ref 625

Elastic APM seems to point to something inside the template:

Image

I suspect this could be due to some select for foreign keys, maybe a missing `raw_id_fields`?

1. Make sure your local DB has 100k+ Partijen
2. Go to the detail page of a Partij
3. Observe that it takes a long time to load

_No response_

stevenbal:
and other admin pages that were missing this as well, to make sure the admin still loads fast if there are a lot of possible options for related objects

Fixes #510 

**Changes**

* Add missing raw_id_fields for PartijAdmin and other admin pages

### DIFF
--- a/src/openklant/components/klantinteracties/admin/partijen.py
+++ b/src/openklant/components/klantinteracties/admin/partijen.py
@@ -126,18 +126,21 @@ class PartijIdentificatorInlineAdmin(admin.StackedInline):
     readonly_fields = ("uuid",)
     model = PartijIdentificator
     form = PartijIdentificatorAdminForm
+    raw_id_fields = ("sub_identificator_van",)
     extra = 0
 
 
 class BetrokkeneInlineAdmin(admin.StackedInline):
     readonly_fields = ("uuid",)
     model = Betrokkene
+    raw_id_fields = ("klantcontact",)
     extra = 0
 
 
 class DigitaalAdresInlineAdmin(admin.StackedInline):
     readonly_fields = ("uuid",)
     model = DigitaalAdres
+    raw_id_fields = ("betrokkene",)
     extra = 0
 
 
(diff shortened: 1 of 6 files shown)


src/openklant/components/klantinteracties/admin/partijen.py
```diff
@@ -299,3 +303,29 @@ class CategorieAdmin(admin.ModelAdmin):
...
+class PartijIdentificatorAdmin(admin.ModelAdmin):
+    readonly_fields = ("uuid",)
+    search_fields = (
+        "partij",
+        "sub_identificator_van",
+    )
+    list_filter = (
+        "partij_identificator_code_objecttype",
+        "partij_identificator_code_soort_object_id",
+        "partij_identificator_object_id",
```
danielmursa-dev, line 318: I think it would be better to remove it from `list_filter` and put it in `search_fields`, otherwise the list would be too large after `n` partij_identificatoren.
```suggestion

```

danielmursa-dev (approved):

Floris272 (commented):
Some admins use autocomplete instead of raw_id fields

- BetrokkeneInlineAdmin.partij
- CategorieRelatieInlineAdmin.categorie
- ParijAdmin.voorkeurs_digitaal_adres


Other Admins like `DigitaalAdresAdmin` &  `BetrokkeAdmin` also use it but are not in the default admin index?

stevenbal:
@Floris272 hmm you're right, though I'm not sure if I want to replace everything with `autocomplete_fields`, because it doesn't allow for finding related objects via filters like `raw_id_fields` does. I think I'll leave it as is for now because it fixes the performance issue, if we get a request to change it to autocomplete we can look into that, but I hope at some point we can replace the entire thing with Open Beheer (though that will probably take some time)

I added the missing admins to `default_admin_index.json`

PR #513, opened 2025-09-26
:bug: [#405] Validate the partijIdentificatie

### LINKED ISSUE #405: Validate the `partijIdentificatie` so that it conforms with `soortPartij`

2.7.0

Initial Issue #345

It is necessary to investigate and then fix the `partijIdentificatie` field when creating or editing the `soortPartij`, to be compliant with the allowed choices, make `partijIdentificatie` also required if necessary.

> if soortPartij is provided in a PATCH request and is set to a different value than the current soortPartij, it should also require the corresponding partijIdentificatie object for that partij soort. I suspect this lack of validation is a problem elsewhere too (e.g. in the admin panel, see https://github.com/maykinmedia/open-klant/issues/303 ). Setting partijIdentificatie to null violates the API spec, which requires this field.

Fixes int the serializer and the admin

_No response_

_No response_

OlhaZahoruiko:
Fixes #405 

**Changes**

[Describe the changes here]

### DIFF
--- a/src/openklant/components/klantinteracties/api/serializers/partijen.py
+++ b/src/openklant/components/klantinteracties/api/serializers/partijen.py
@@ -482,6 +482,10 @@ class PartijSerializer(NestedGegevensGroepMixin, PolymorphicSerializer):
         },
         same_model=False,
         group_field="partij_identificatie",
+        help_text=_(
+            "De toegestane eigenschappen zijn afhankelijk van `soortPartij`. "
+            "Onbekende eigenschappen zijn niet toegestaan."
+        ),
     )
     betrokkenen = BetrokkeneForeignKeySerializer(
         read_only=True,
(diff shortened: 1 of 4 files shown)


src/openklant/components/klantinteracties/api/serializers/partijen.py
```diff
@@ -664,6 +664,19 @@ def update_or_create_partij_identificator(self, partij_identificator):
     def update(self, instance, validated_data):
         method = self.context.get("request").method
         partij_identificatie = validated_data.pop("partij_identificatie", None)
+        soort_changed = (
+            "soort_partij" in validated_data
+            and validated_data["soort_partij"] != instance.soort_partij
+        )
+
+        if soort_changed and partij_identificatie is None:
```
stevenbal: Just adding this won't really solve the issue, because essentially the current behavior is that the value of `partijIdentificatie` gets reset to `null` if you change the `soortPartij`, which is not ideal.

I think we should make sure that the shape of `partijIdentificatie` is validated according to the value of `soortPartij`, it seems though that currently the Open Klant APIs ignore extra fields in request bodies (e.g. if I do a POST with `someNonExistentField`: "foo", Open Klant just ignores it) and the same seems to happen to fields nested inside `partijIdentificatie`.

Could you check how Open Zaak handles extra fields (does it raise an error or does it ignore them)?

stevenbal (changes requested):

src/openklant/components/klantinteracties/api/serializers/partijen.py
OlhaZahoruiko: I checked Open Zaak and it behaves the same way: unknown (including nested) fields are ignored. That means the current serializer won't reject a persoon-shaped partijIdentificatie for an organisatie. Would you prefer fixing this in the nested serializers or by adding explicit validation in PartijSerializer?

OlhaZahoruiko: Created a separate issue in commonground-api-common for adding optional strict shape validation to `PolymorphicSerializer` 
https://github.com/orgs/maykinmedia/projects/15/views/16?pane=issue&itemId=213532889&issue=maykinmedia%7Ccommonground-api-common%7C170

stevenbal:
I'm going to leave this on hold until https://github.com/maykinmedia/commonground-api-common/issues/170 is discussed during refinement

PR #530, opened 2025-10-20
:sparkles: [#500] Implement Referentielijsten integration for Klantcontact

### LINKED ISSUE #500: Gebruikmaken van de Referentielijsten API voor het valideren van de gebruikte kanalen voor klantcontacten
### Thema / Theme

Klantinteracties API

Momenteel dien je bij een nieuw /klantcontact verplicht een kanaal op te geven.

Dit veld is van wat ik heb gezien een string (max 50 charts) zonder additionele validatie hierop. Je kan dus vrij makkelijk vervuiling krijgen doordat verschillende applicaties verschillende termen gebruiken voor kanalen (eg: 'email' vs 'e-mail', of 'website' vs 'web' vs 'utrecht.nl'). Op zich geen probleem voor de eerste of tweede applicatie op Open Klant maar daarna wordt het makkelijk een rommeltje aangezien er geen validatie verder plaatsvindt op dit veld

@frankmijn 's voorstel is om Open Klant uit te breiden met een integratie met de Referentielijsten API, waarbij indien gekoppeld het gebruikte kanaal wordt gevalideerd op de lijst met 'kanalen' in deze API.

Mocht bij het aanmaken van een /klantcontact het gebruikte kanaal niet aanwezig zijn dan mag het klantcontact niet worden aangemaakt. Het zou dan wel chique zijn om de wel-aanwezige kanalen terug te keren in de foutmelding naar de aanroeper.

Heeft een relatie met Open Product issue: https://github.com/maykinmedia/open-product/issues/111 

_No response_

_No response_

OlhaZahoruiko:
Fixes #500

**Changes**
- [ ] **Add ReferentielijstenConfig singletonmodel**
  - [ ] Add `enabled` field (default false)
  - [ ] Add `service` field
  - [ ] Add `tabel_code` field (CharField to indicate which tabel contains the Kanalen)
  - [ ] Display-only: show at least the HTTP status code (preferably also the list of items) when performing a GET with this `tabel__code` (similar to `zgw_consumers.Service` status check)
  - [ ] On model save, validate that all existing `Klantcontact.kanaal` in the database are present in the linked tabel

- [ ] **Create a ReferentielijstenClient** in `src/referentielijsten/client.py` that subclasses `zgw_consumers.NLXClient`
  - [ ] Add a method `get_items_by_tabel_code` that does a GET on `/items?tabel__code={code}`
  - [ ] Add another method that calls the above method and caches the result for a configurable amount of time (default 5 minutes)

- [ ] **Add a validator to Klantcontact**
  - [ ] Return early if `ReferentielijstenConfig.enabled == False`
  - [ ] Instantiate a `ReferentielijstenClient` based on `ReferentielijstenConfig.service`
  - [ ] Call the cached version of `get_items_by_tabel_code` and raise a `ValidationError` if the supplied `Kanaal` is not found (validation error should list all valid kanalen)

- [ ] **Add a configuration step** to load this config via `setup_configuration`

- [ ] **Add a page in the docs** under Manual explaining this optional integration


src/openklant/config/tests/test_models.py
```diff
@@ -0,0 +1,57 @@
...
+
+from openklant.components.klantinteracties.api.validators import KanaalValidator
+from openklant.components.klantinteracties.models import Klantcontact
+from openklant.config.models import ReferentielijstenConfig
+
+
+class ReferentielijstenConfigTestCase(TestCase):
+    @classmethod
+    def setUpTestData(cls):
+        service, _ = Service.objects.update_or_create(
```
danielmursa-dev: Also here: `from zgw_consumers.test.factories import ServiceFactory`

src/openklant/config/models.py
```diff
@@ -0,0 +1,50 @@
...
+
+from solo.models import SingletonModel
+from zgw_consumers.models import Service
+
+from openklant.components.klantinteracties.models import Klantcontact
+from referentielijsten_client.client import ReferentielijstenClient
+
+
+class ReferentielijstenConfig(SingletonModel):
+    enabled = models.BooleanField(default=False)
```
danielmursa-dev:
```suggestion
    enabled = models.BooleanField(default=False, help_text="Indicates if Referentielijsten is enabled")
```
but i think should be in Dutch, with translations, same for other `help_text`

src/openklant/config/models.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def clean(self):
+        if not self.enabled:
+            return
+
+        if not self.service or not self.tabel_code:
+            raise ValidationError(
+                "Service and tabel_code must be set when validation is enabled."
+            )
+
```
danielmursa-dev: If these fields are mandatory, so you can simply remove `blank=True`, and `null=True` from the fields, in this way they will become required

src/openklant/config/models.py
```diff
@@ -0,0 +1,50 @@
...
+        client = ReferentielijstenClient(service=self.service)
+        items = client.get_items_by_tabel_code(self.tabel_code)
+        valid_kanalen = {item.get("naam") for item in items if item.get("naam")}
```
danielmursa-dev:
```suggestion
        valid_kanalen = {item.get("code") for item in items if item.get("code")}
```
Is better to check the code, which is unique, i think also in the `KanaalValidator` you check the `code`, right?

OlhaZahoruiko: yes

src/referentielijsten_client/steps.py
```diff
@@ -0,0 +1,61 @@
...
+from django_setup_configuration.exceptions import ConfigurationRunFailed
+from zgw_consumers.models import Service
+
+from openklant.config.models import ReferentielijstenConfig
+from referentielijsten_client.models import ReferentielijstenConfigurationModel
+
+logger = structlog.stdlib.get_logger(__name__)
+
+
+class ReferentielijstenConfigurationStep(
```
stevenbal: Can move this to `src/referentielijsten_client/setup_configuration/steps.py` and also add `src/referentielijsten_client/setup_configuration/tests/test_client_config_step.py` to make sure this step works as expected? You can see an example of how tests are made for this in `src/openklant/setup_configuration/tests/test_token_auth_config.py`

I think the following two tests are the only necessary ones:
- happy flow where the `selectielijst_api_service_identifier` points to an existing Service in the database -> assert that the config is created with the configured values
- errors should be raised if the `selectielijst_api_service_identifier` does not point to an existing Service
- running the same step twice should be idempotent

src/referentielijsten_client/steps.py
```diff
@@ -0,0 +1,61 @@
...
+        except ValidationError as exc:
+            logger.error("configuration_validation_failed", exc_info=exc)
+            raise ConfigurationRunFailed(
+                f"Validation failed while configuring Referentielijsten: {exc}"
+            ) from exc
+        except Exception as exc:
+            logger.error("configuration_failed", exc_info=exc)
+            raise ConfigurationRunFailed(
+                f"Failed to configure Referentielijsten: {exc}"
+            ) from exc
```
stevenbal: I think the try-except block can be removed if the `full_clean` is removed as well
```suggestion
```

src/referentielijsten_client/models.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class ReferentielijstenConfigurationModel(models.Model):
+    enabled = models.BooleanField(
+        default=False,
+        help_text=_("Indicates whether Referentielijsten is enabled"),
+    )
+    service_api_root = models.URLField(
+        help_text=_("The API root URL of the Referentielijsten service")
+    )
```
stevenbal:
```suggestion
    referentielijsten_api_service_identifier: str = DjangoModelRef(
        ReferentielijstenConfig,
        "service",
        examples=["referentielijsten-api"],
        default=...,
    )
```

src/openklant/config/models.py
```diff
@@ -0,0 +1,73 @@
...
+from zgw_consumers.client import build_client
+from zgw_consumers.models import Service
+
+from openklant.components.klantinteracties.models import Klantcontact
+
+
+class ReferentielijstenConfig(SingletonModel):
+    enabled = models.BooleanField(
+        default=False,
+        help_text=_("Geef aan of Referentielijsten is ingeschakeld"),
```
stevenbal:
```suggestion
        help_text=_("Geef aan of de optionele integratie met Referentielijsten API is ingeschakeld"),
```

src/openklant/config/tests/test_models.py
```diff
@@ -0,0 +1,148 @@
...
+                "tabel": {"code": MAPPED_TABLE_CODE},
+                "aanvullendeGegevens": item_fields.get("aanvullende_gegevens"),
+            }
+        )
+
+my_vcr = vcr.VCR(
+    cassette_library_dir="tests/cassettes",
+    record_mode="once",
+    match_on=["method", "url", "query"],
+)
```
stevenbal: I think we should go over how VCR works if we're both in the office thursday (or we could do it via slack), because the current setup for these tests do not actually use VCR the way it's intended.

The current setup for the tests still relies on a lot of mocking via `unittest.mock.patch` for the client, which is prone to errors and assumptions that might not be true in practice (misspelled method names, etc.)

With VCR the setup is like this:
- instead of mocking, we make sure that we have a real docker-compose setup for the service we want to connect to (which is what the `docker-compose.referentielijsten.yml` is for), at `http://localhost:8004/api/v1/` in this case
- in the tests, we make sure to configure the `ReferentelijstenConfig` to point to this real Referentielijsten instance
- the tests will make actual network calls to this instance (no mocking)
- VCR records the responses that the real service returns and stores them in cassettes (yaml files)
- because VCR recorded the responses, we don't need to have the `docker-compose.referentielijsten.yml` running for subsequent test runs, because VCR will return the response that was recorded in the cassettes

The VCRTestCase from `maykin-common` should do most of this for you: https://maykin-django-common.readthedocs.io/en/latest/reference/vcr.html#maykin_common.vcr.VCRTestCase, generally you don't have to override the TEST_FILES directory

docker/setup_configuration/data.yaml
```diff
@@ -64,9 +64,24 @@ zgw_consumers:
...
       secret: open-klant-secret
 
+    - identifier: referentielijsten-api
+      label: Referentielijsten API
+      api_root: http://referentielijsten-web.local:8000/api/v1/
+      api_type: brc
+      auth_type: zgw
+      client_id: open-klant-referentielijsten
+      user_id: open-klant-referentielijsten
+      secret: open-klant-referentielijsten-secret
```
stevenbal: Referentielijsten doesn't have any API authentication/authorization, so this can be left empty. The API type is technically "orc" ("overige"), but it doesn't really matter in practice afaik
```suggestion
      api_type: orc
      auth_type: no_auth
```

src/openklant/components/klantinteracties/api/serializers/klantcontacten.py
```diff
@@ -332,6 +339,12 @@ class Meta:
                 "lookup_field": "uuid",
                 "help_text": _("De unieke URL van dit klantcontact binnen deze API."),
             },
+            "kanaal": {
+                "help_text": _(
+                    "Communicatiekanaal dat bij het klantcontact werd gebruikt."
+                ),
```
stevenbal: This should be automatically picked up from the model field
```suggestion
```

src/openklant/components/klantinteracties/api/serializers/klantcontacten.py
```diff
@@ -306,6 +307,12 @@ class KlantcontactSerializer(serializers.HyperlinkedModelSerializer):
         "had_betrokkenen.digitale_adressen": f"{SERIALIZER_PATH}.digitaal_adres.DigitaalAdresSerializer",
     }
 
+    kanaal = serializers.CharField(
+        max_length=50,
+        help_text=_("Communicatiekanaal dat bij het klantcontact werd gebruikt."),
+        validators=[KanaalValidator()],
+    )
```
stevenbal:
```suggestion
```

docker/docker-compose.referentielijsten.yml
```diff
@@ -0,0 +1,55 @@
...
+      - IS_HTTPS=no
+      - CACHE_DEFAULT=referentielijsten-redis:6379/0
+      - CACHE_AXES=referentielijsten-redis:6379/0
+      - SUBPATH=${SUBPATH:-/}
+      - DJANGO_SUPERUSER_PASSWORD=admin
+    ports:
+      - 8004:8000
+    volumes: &referentielijsten_web_volumes
+      # mount fixtures dir to automatically populate the DB
+      - ./fixtures/:/app/fixtures
```
stevenbal:
```suggestion
      - ./referentielijsten/fixtures/:/app/fixtures
```

docker/referentielijsten/fixtures/referentielijsten_fixtures.json
```diff
@@ -0,0 +1,39 @@
+[
+{
+    "model": "api.tabel",
```
stevenbal: Could you also add a superuser to this fixture? That makes it a bit easier to update/extend this fixture later on

src/referentielijsten_client/tests/test_client.py
```diff
@@ -0,0 +1,55 @@
...
+
+        self.assertIsInstance(items, list)
+        self.assertTrue(all("code" in item for item in items))
+        codes = [item["code"] for item in items]
+        self.assertIn("email", codes)
+        self.assertIn("phone", codes)
+
+    def test_get_cached_items_by_tabel_code(self):
+        first = self._get_items_from_fixture("KANAAL")
+        second = self._get_items_from_fixture("KANAAL")
```
stevenbal: It would be nice if this test also verifies that only one network call happens (because it's cached by the underlying method). I don't fully remember how to check this, but I think maybe with something like `self.vcr._current_cassette.requests` you could check this

src/openklant/config/tests/test_models.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def test_validation_fails_for_invalid_kanaal(self):
+        validator = KanaalValidator()
+
+        with self.assertRaises(ValidationError) as cm:
+            validator("fax")
+
+        self.assertIn("'fax' is not a valid kanaal", str(cm.exception))
+        self.assertIn("email", str(cm.exception))
+        self.assertIn("phone", str(cm.exception))
```
stevenbal: Instead of testing this by calling the validator directly, could you test this by making API calls (POST/PUT/PATCH) to `/klantcontacten`?

I think it would also be good to have a test for the behavior of the API if:
1. referentielijsten returns a 400 or 500 error
2. a requestexception or timeout occurs (you could use https://maykin-django-common.readthedocs.io/en/latest/reference/vcr.html#maykin_common.vcr.VCRMixin.vcr_raises for this)

stevenbal (changes requested):
@OlhaZahoruiko there are two remaining checkbox items in the original issue (about documentation and showing the response code in the config model admin), but since this PR is quite big already let's put those in a separate PR when this one is merged

stevenbal:
@OlhaZahoruiko can you pick up the remaining points from the checklist in a separate PR?

PR #541, opened 2025-11-20
:sparkles: [#500] display HTTP status and items for tabel__code GET checks

### LINKED ISSUE #500: Gebruikmaken van de Referentielijsten API voor het valideren van de gebruikte kanalen voor klantcontacten

OlhaZahoruiko:
Fixes #500

**Changes**

[Describe the changes here]


src/openklant/components/klantinteracties/api/tests/test_kanaal_referentielijsten.py
```diff
@@ -238,3 +238,58 @@ def test_put_with_invalid_kanaal(self, mock_get_solo):
             error["reason"],
             "'carrier-pigeon' is not a valid kanaal. Allowed values: phone, email",
         )
+
+    @patch(
+        "openklant.components.klantinteracties.api.validators.ReferentielijstenConfig.get_solo"
+    )
```
danielmursa-dev: Here you can add `@freeze_time("2025-01-01T12:00:00Z")` in this way when you test the `datetime.now()` the date is fixed

src/openklant/components/klantinteracties/api/validators.py
```diff
@@ -235,6 +233,33 @@ def __call__(self, value: str):
...
+            )
+            eind_dt = (
+                datetime.fromisoformat(eind.replace("Z", "+00:00")) if eind else None
+            )
+
+            if begin_dt and now < begin_dt:
+                return False
+            if eind_dt and now >= eind_dt:
+                return False
+            return True
```
danielmursa-dev: This is more easy to understand i think, and is the same like here?
`Only consider items valid if the current timestamp >= beginGeldigheid and < eindGeldigheid`
```suggestion
            if begin_dt and eind_dt:
                return begin_dt <= now < eind_dt
            return True
```

however, i am not sure if the validation is still valid if the dates are None. ?!, i don't know.

Add test also for these cases

src/openklant/config/tests/test_admin.py
```diff
@@ -182,3 +182,25 @@ def test_existing_klantcontacten_with_invalid_kanalen_raises_error(self):
...
+        form["enabled"] = True
+        form["service"] = self.service.pk
+        form["kanalen_tabel_code"] = "KANAAL"
+
+        form.submit()
+
+        config = ReferentielijstenConfig.get_solo()
+
+        try:
+            result = config.status_check()
```
stevenbal: This is a general point: tests that only test internal methods like this are not enough to test functionality. Currently this `status_check` method is not used anywhere except in this test, so if I were to merge this now, this status check functionality would not be part of the admin at all. This is also why I think it's important to not just rely on unittests while developing something, but to also manually check if it works in the admin

There should at least be one test on the admin view to verify that this method is actually called there, that way we know that the functionality is actually used

stevenbal (changes requested):

PR #548, opened 2025-12-01
:sparkles: [#544] Send nl.overheid.zaken.zaak-gelinkt event

### LINKED ISSUE #544: Support Open Archiefbeheer destruction in Open Klant
- [ ] Be able to recognise when a Zaak is linked, in an unambiguous way.
   Best we can do here is check on 
   ```
   Onderwerpbjectintificatiecator.codeObjecttype == "zaak"
                                 .codeSoortObjectId == "uuid"
   ```
 - [x] Send `nl.overheid.zaken.zaak-gelinkt` events after successful POSTs that link     Onderwerpobject in the registry to a Zaak. See https://github.com/open-zaak/open-zaak/issues/2251
   Because we don't have a URL, `zaak` field in the event data will be the URN `f"urn:uuid:{Onderwerpbjectintificatiecator.objectId}"`

```
Request: DELETE: /api/v1/onderwerpobjecten/{uuid}?cascade=True
Response: 
     204 when gone (deleted with cascasde)
     200 when deleted (without cascade)  Respons content list[url] of the Klantcontact that weren't deleted. Then the OAB plugin can decide for itself what to do with it.
```

Details:

When deleting an OnderwerpObject with cascade=True, the ENDPOINT should take care of removing the  Klantcontact and Betrokkene.

- [x] Questions: do we EVER unlink?

  Answer: Yes. The Foreign Key is from Onderwerpobject to 2 Klantcontact. So in theory it is possible that other Onderwerpobject are linked to the same Klantcontact. If a Klantcontact has multiple `Onderwerpobject`, we only delete this current Onderwerpobject and don't delete that Klantcontact.

Caution: It may seem faster to cascade all if those Onderwerpobject point to the same zaak, but since there are 2 Klantcontact connected to a Onderwerpobject, cascading them, may leave dangling/zombie Klantcontacten, unreachable from the Zaak. So a LEFT JOIN with a recursive CTE query would be needed.

**WE MUST NEVER CASCADE DELETE PARTIJ** the rest, like Bijlagen etc, should cascade.

**LABEL OF THE CLOUD EVENT**

klantcontact__ onderwerp of either klantcontact or was_klantcontact, (truncated to 80 chars)

OlhaZahoruiko:
Fixes #544

**Changes**

Send nl.overheid.zaken.zaak-gelinkt events after successful POSTs that link Onderwerpobject in the registry to a Zaak. See
https://github.com/open-zaak/open-zaak/issues/2251
Because we don't have a URL, zaak field in the event data will be the URN f"urn:uuid:{Onderwerpbjectintificatiecator.objectId}"

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -254,7 +254,7 @@ mozilla-django-oidc-db==1.1.1
     # via
     #   -r requirements/base.in
     #   open-api-framework
-notifications-api-common==0.10.0
+notifications-api-common==0.10.1
     # via commonground-api-common
 open-api-framework==0.13.2
     # via -r requirements/base.in
--- a/requirements/ci.txt
+++ b/requirements/ci.txt
@@ -496,7 +496,7 @@ mozilla-django-oidc-db==1.1.1
     #   open-api-framework
 multidict==6.0.5
     # via yarl
-notifications-api-common==0.10.0
+notifications-api-common==0.10.1
     # via
     #   -c requirements/base.txt
     #   -r requirements/base.txt
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -609,7 +609,7 @@ multidict==6.0.5
     #   yarl
 nodeenv==1.9.1
     # via pre-commit
-notifications-api-common==0.10.0
+notifications-api-common==0.10.1
     # via
     #   -c requirements/ci.txt
     #   -r requirements/ci.txt


src/openklant/components/klantinteracties/api/viewsets/klantcontacten.py
```diff
@@ -331,6 +332,20 @@ def perform_create(self, serializer):
             token_identifier=token_auth.identifier,
             token_application=token_auth.application,
         )
+        object_type = instance.onderwerpobjectidentificator.get("code_objecttype")
+        if object_type == "zaak":
```
stevenbal: Could you add the setting `ENABLE_CLOUD_EVENTS` (same as https://github.com/open-zaak/open-zaak/blob/2be1bc48a93107205a9148f2899bc33f810fd7fa/src/openzaak/conf/includes/base.py#L531-L537) and make sure the event is only sent if this setting is True?

You can remove the `add_to_docs=False` part

src/openklant/components/klantinteracties/api/viewsets/klantcontacten.py
```diff
@@ -331,6 +332,20 @@ def perform_create(self, serializer):
...
+            process_cloudevent(
```
stevenbal: I noticed that the `send_cloudevent` from `notifications-api-common` currently uses the `Content-Type: application/json` header, while the `/cloudevents` endpoint requires the `application/cloudevents+json` content-type, so if I test this locally with Open Notificaties, I get HTTP 415 error codes. Could you make sure this is changed in the library to use the `application/cloudevents+json` content-type header?

src/openklant/components/klantinteracties/api/viewsets/klantcontacten.py
```diff
@@ -331,6 +332,20 @@ def perform_create(self, serializer):
...
+                type="nl.overheid.zaken.zaak-gelinkt",
+                subject=instance.onderwerpobjectidentificator.get("object_id"),
+                data={
+                    "zaak": f"urn:uuid:{instance.onderwerpobjectidentificator.get('object_id')}",
+                    "linkTo": f"urn:uuid:{instance.uuid}",
```
stevenbal: I think for `linkTo` we'll have to use the URL of the OnderwerpObject, because Open Zaak requires a URL to create the ZaakObject: 
afbeelding

@CharString is this what you had in mind when you specced https://github.com/open-zaak/open-zaak/issues/2251? Since you didn't explicitly mention `linkTo` should be a URL

src/openklant/components/klantinteracties/api/viewsets/klantcontacten.py
```diff
@@ -331,6 +332,20 @@ def perform_create(self, serializer):
             token_identifier=token_auth.identifier,
             token_application=token_auth.application,
         )
+        object_type = instance.onderwerpobjectidentificator.get("code_objecttype")
+        if object_type == "zaak":
+            process_cloudevent(
+                type="nl.overheid.zaken.zaak-gelinkt",
```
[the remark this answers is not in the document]
OlhaZahoruiko: add a new section called Functionality, with a separate page for Cloud events

docs/functionality/cloud_events.rst
```diff
@@ -0,0 +1,46 @@
...
+            "linkTo": "/klantinteracties/api/v1/onderwerpobjecten/12345678-aaaa-bbbb-cccc-999999999999",
+            "label": "Mijn Klantcontact Onderwerp",
+            "linkObjectType": "Onderwerpobject"
+        }
+    }
+
+The shape of these cloud events and the actions that trigger them are still subject to
+change. Currently, cloud events are delivered directly to the configured endpoint of
+Open Notificaties, but in the future it will be possible to route these cloud events
+using Open Notificaties message routing as well.
```
stevenbal: These cloud events are already sent to Open Notificaties, so I think this disclaimer can be removed. This is currently only applicable for Open Zaak, because cloud events there are sent directly to Logius instead of Open Notificaties
```suggestion
change.
```

docs/functionality/cloud_events.rst
```diff
@@ -0,0 +1,44 @@
...
+
+Cloud events
+------------
+
+Open Klant can emit cloud events when configured to do so via the
+``ENABLE_CLOUD_EVENTS`` setting.
+
+Currently, Open Klant emits the following cloud event:
+
+* ``zaak-gelinkt``: emitted when an ``Onderwerpobject`` is created that links to a
```
stevenbal:
```suggestion
* ``nl.overheid.zaken.zaak-gelinkt``: emitted when an ``Onderwerpobject`` is created that links to a
```

src/openklant/components/klantinteracties/api/viewsets/klantcontacten.py
```diff
@@ -331,6 +332,20 @@ def perform_create(self, serializer):
...
+                    "linkTo": f"urn:uuid:{instance.uuid}",
+                    "label": str(instance.klantcontact)
+                    if instance.klantcontact
+                    else "Onderwerpobject",
+                    "linkObjectType": "Onderwerpobject",
+                },
+            )
 
     @transaction.atomic
     def perform_update(self, serializer):
```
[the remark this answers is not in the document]
stevenbal: I suggested changing the names to `zaak-gekoppeld` and `zaak-ontkoppeld` (https://github.com/open-zaak/open-zaak/pull/2270/files#r2618847654), but I won't let it be blocking for this PR, if we end up changing it, we can do it in a separate PR

PR #592, opened 2026-03-31
:sparkles: [#529] Add filtering by subIdentificatorVan in partijen en…

### LINKED ISSUE #529: Additional filtering using 'subIdentificatorVan' data to retrieve partijen to be able to retrieve a vestiging partij for a certain related KVK number
### Thema / Theme

Klantinteracties API

A vestigingsnummer alone is not a unique identifier for a vestiging. The KVK number of the related company always needs to be provided as well. If only because vestigingen can change ownership over time. In such a case the vestigingsnummer remains the same but the related KVK number changes. It is therefore important that all vestiging information in Open Klant is always retrieved using both the vestigingsnummer as well as the KVK number. For example contact details (digital addresses) of a vestiging are likely to be different depending on which KVK company a vestiging is linked too. And it is perfectly normal if there are multiple ('historical') vestigingen in Open Klant with the same vestigingsnummer but with a different KVK number. 

The requested feature would be a change in the 'partijen lijst' API endpoint. Currently you can only filter on the `partijIdentificator` itself (using `partijIdentificator__objectId` and other parameters) but for the use case above we would also want to be able to filter on the `subIdentificatorVan` (e.g. using parameters like `subPartijIdentificatorVan__objectId` etc).

With this feature we no longer need to do non-trivial filtering on the client side to get the correct vestiging partij which is linked to the correct parent company partij with the correct KVK number. This reduces client-side complexity for all Open Klant clients that need this functionality.

For background in the context of PodiumD regarding KVK koppelvlak: https://dimpact.atlassian.net/wiki/spaces/POO/pages/393969665/KvK+koppelvlak

For background for our (ZAC) specific use case, see: https://dimpact.atlassian.net/browse/PZ-8029

OlhaZahoruiko:
…dpoint

Fixes #529 

**Changes**

[Describe the changes here]


stevenbal (changes requested):

src/openklant/components/klantinteracties/api/filterset/partijen.py
```diff
@@ -130,6 +158,38 @@ def filter_categorierelatie_categorie_naam(self, queryset, name, value):
         except ValueError:
             return queryset.none()
 
+    def filter_sub_identificator_object_id(self, queryset, name, value):
+        try:
+            return queryset.filter(
+                partijidentificator__sub_identificator_van__partij_identificator_object_id=value
```
stevenbal: The same issue I mentioned earlier still applies here: by filtering on ``sub_identificator_van``, the filters are still applied to the parent identifier, because `sub_identificator_van` points to the parent.

You can filter on the related_name `parent_partij_identificator` (which is incorrectly named actually, because it should be `sub_partij_identificator`. Could you change that name as well?) https://github.com/maykinmedia/open-klant/blob/3d6848afabc9ce6ff91a72dc41333f2447739e5b/src/openklant/components/klantinteracties/models/partijen.py#L319

```suggestion
                partijidentificator__sub_partij_identificator__partij_identificator_object_id=value
```

This applies to the other filters as well

stevenbal (changes requested):

stevenbal:
@OlhaZahoruiko sorry, I was wrong in my first PR feedback :facepalm: 

I was confused about the usage of `subIdentificatorVan__` filters and `partijIdentificator__`, I think your initial version was correct, so I changed it back to that so that you can find a specific branch of a company by:

* partijIdentificator parameters: the branch specific info
* subIdentificatorVan: the info related to the KVK number

stevenbal (approved):

PR #600, opened 2026-04-24
Issue/593 empty partij identificator

### LINKED ISSUE #593: Creating an "empty" PartijIdentificator blocks any future PartijIdentificator creation

2.15.0

Ran into this while reviewing an unrelated PR locally:

Apparently it is possible to do a POST on `/klantinteracties/api/v1/partij-identificatoren` with an "empty" PartijIdentificator by either specifying an empty JSON body `{}` or no body at all even, and Open Klant will respond with a 201. After this however, any POST to this same endpoint any body will result in a 400 error:

```json
{
  "type": "http://localhost:8000/ref/fouten/ValidationError/",
  "code": "invalid",
  "title": "Ongeldige invoerwaarde.",
  "status": 400,
  "detail": "Ongeldige invoerwaarde.",
  "instance": "urn:uuid:22a7b2c9-1ffd-4c47-8851-e1d0cc9e8264",
  "invalidParams": [
    {
      "name": "__all__",
      "code": "invalid",
      "reason": "`PartijIdentificator` moet uniek zijn, er bestaat er al een met deze gegevenscombinatie."
    }
  ]
}
```

And this is not resolved until the "empty" PartijIdentificator is deleted

1. Do a POST on `/klantinteracties/api/v1/partij-identificatoren` without a body
2. Try to do a POST again but with a valid body
3. Observe the 400 error

In any case, having this empty PartijIdentificator should not block all subsequent PartijIdentificator from being created.

Additionally, I'm not sure what the use case is for an empty Partijidentificator, so perhaps we should disallow that from being created as well

TimdeBeer1:
Fixes #593 

**Changes**

Added validation to make partijIdentificator required.

### DIFF
--- a/src/openklant/components/klantinteracties/admin/partijen.py
+++ b/src/openklant/components/klantinteracties/admin/partijen.py
@@ -1,5 +1,6 @@
 from django import forms
 from django.contrib import admin
+from django.core.exceptions import ValidationError
 from django.utils.translation import gettext_lazy as _
 
 from openklant.components.klantinteracties.models.rekeningnummers import Rekeningnummer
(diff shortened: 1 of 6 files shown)


src/openklant/components/klantinteracties/api/serializers/partijen.py
```diff
@@ -398,8 +398,6 @@ class PartijIdentificatorSerializer(
         source="partij",
     )
     partij_identificator = PartijIdentificatorGroepTypeSerializer(
-        required=False,
-        allow_null=True,
```
stevenbal, line 402: Let's explicitly set `required=True` and `allow_null=False` to make it clear

src/openklant/components/klantinteracties/admin/partijen.py
```diff
@@ -83,12 +84,30 @@ def clean(self):
 
 
 class PartijIdentificatorAdminForm(forms.ModelForm):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields["partij_identificator_code_objecttype"].required = True
+        self.fields["partij_identificator_code_soort_object_id"].required = True
+        self.fields["partij_identificator_object_id"].required = True
+        self.fields["partij_identificator_code_register"].required = True
```
stevenbal, line 92: I think both this and the added code in `clean` can be removed, if `blank=True` is removed from the model fields `partij_identificator_code_objecttype`, etc., because forms derive required/not required from the value of `blank` on the model fields

stevenbal (changes requested):

src/openklant/components/klantinteracties/admin/partijen.py
TimdeBeer1: I tried removing the blank on the model fields, after doing that making a post i get a invalidParams with fields can't be empty while having valid data in it. I thought it was something about how the grouping works

stevenbal: Hmm I see, sort of surprised that happen, maybe the serializer calls `clean` which then causes `blank=False` to be enforced, probably not trivial to fix, so let's keep it as is

stevenbal (approved):

PR #608, opened 2026-06-03
Feature/598 contactmoment notificatie

### LINKED ISSUE #598: Notificaties op contactmomenten
### Thema / Theme

Klantinteracties API

Result of technische stuurgroep: https://www.notion.so/platformvoordienstverlening/Klantcontact-notificaties-33378b5f4db4802a9782d71dde656c80

Request is to add notifications to the Notifications API for klantcontacten.
To properly rotate the (failed) klantcontacten the following data needs to be added to the klantcontact:
1. `Type` - Description: Defines the -type of the subject of the contact. Used to define origin. Can be Zaak/Product/bericht -Type
    1. What reference should this be? URL? URN?
2. `IndicatieVerdereActieOndernomen` (Please propose a better fitting name 😄) - Description: Used in case of a failed klantcontact. Defines if the system which created the klantcontact will undertake further remediation and thus this klantcontact can be ignored by other systems. (A subsequent success of failed klantcontact will come)
    1. So in case a contacting system will have operators trying to call back later, or a messaging service will after this failed email try SMS next

The `kenmerken` of the notification should contain (at least, maybe more would be useful?):
1. Existing field `indicatieContactGelukt`
2. `Type` as described above
3. `IndicatieVerdereActieOndernomen` as described above
4. To discuss: Include onderwerpObjecten?
    1. This could make the notification quite big but could be nice to immediately define the corresponding zaak / product / ... to rotate to the right sub process/dossier. 
    2. This could also be left to the ZAC/TSA after retrieving the klantcontact + onderwerpObjecten

HoofdObject & Resource URL should be: the URL to the klantcontact

Example of the notification:
```
{
    "actie": "create",
    "kanaal": "klanten",
    "resource": "klantcontacten",
    "kenmerken": {
        "type": "https://openzaak-zgw.acc.denhaag.nl/catalogi/api/v1/zaaktypen/b51b2a95-36ab-4628-8dbe-c2ecabc23afa",
		"indicatieContactGelukt": false,
		"OnderwerpObjecten": { "To be refined": true },
		"IndicatieVerdereActieOndernomen": false
    },
    "hoofdObject": "{KlantenURL}",
    "resourceUrl": "{KlantenURL}",
    "aanmaakdatum": "2026-04-13T09:44:04.127Z"
}

```

ZACs can take remediation actions based upon failed klantcontacten

_No response_

Floris272:
Fixes #598 

**Changes**
- Adds verdere_actie_ondernomen & hoofd_onderwerp_type to klantcontact
- Adds klantcontact kanaal
- Adds klantcontact notifications

[Describe the changes here]

**Checklist**

- Commit hygiene

  - [ ] Commit messages refer to the relevant Github issue
  - [ ] Commit messages explain the "why" of change, not the how

- Architectural design record

  - [ ] If a design decision was made that changes functionality, create an architectural design record for it [here](https://github.com/maykinmedia/open-klant/wiki/Architectural-Decision-Records-(ADR))


src/openklant/components/klantinteracties/models/klantcontacten.py
```diff
@@ -82,6 +82,21 @@ class Klantcontact(APIMixin, models.Model):
...
+        blank=True,
+    )
+
+    verdere_actie_ondernomen = models.BooleanField(
+        _("verdere actie ondernomen"),
+        help_text=_(
+            "Geeft aan of het systeem waar het contact heeft plaats gevonden zelf actie onderneemt."
+        ),
+        null=True,
+        blank=True,
```
danielmursa-dev: Wouldn't it be better to set `default=False`?

src/openklant/components/klantinteracties/models/klantcontacten.py
```diff
@@ -82,6 +82,21 @@ class Klantcontact(APIMixin, models.Model):
         ),
         null=True,
     )
+    hoofd_onderwerp_type = models.URLField(
+        _("hoofd onderwerp type"),
+        help_text=_("Het type van het onderwerp van het contact"),
+        null=True,
```
stevenbal: I don't think we have to make this nullable, right? Just specifying blank=True should be enough to make it optional via the API and allow empty strings

```suggestion
```

src/openklant/components/klantinteracties/models/klantcontacten.py
```diff
@@ -82,6 +82,21 @@ class Klantcontact(APIMixin, models.Model):
...
+        blank=True,
+    )
+
+    verdere_actie_ondernomen = models.BooleanField(
+        _("verdere actie ondernomen"),
+        help_text=_(
+            "Geeft aan of het systeem waar het contact heeft plaats gevonden zelf actie onderneemt."
```
stevenbal:
```suggestion
            "Geeft aan of het systeem waar het contact heeft plaats gevonden zelf actie onderneemt als het niet lukt om de klant te bereiken."
```

src/openklant/components/klantinteracties/models/klantcontacten.py
stevenbal: Yeah I'd say we remove null=True and set the default to False

src/openklant/components/klantinteracties/models/klantcontacten.py
```diff
@@ -82,6 +82,21 @@ class Klantcontact(APIMixin, models.Model):
...
         null=True,
     )
+    hoofd_onderwerp_type = models.URLField(
+        _("hoofd onderwerp type"),
+        help_text=_("Het type van het onderwerp van het contact"),
+        null=True,
+        blank=True,
+    )
+
+    verdere_actie_ondernomen = models.BooleanField(
```
stevenbal, line 94: I'm not really sure about this name, because it's in present perfect and reads as though further action has already been taken, but it's to indicate that further action will be taken if necessary. I couldn't really come up with anything better though, `verdereActieWordtOndernomen` would be more correct but it also turns into a bit of a sentence then.

@Floris272 do you have any ideas?

src/openklant/components/klantinteracties/kanalen.py
```diff
@@ -23,3 +24,14 @@
         "status",
     ),
 )
+
+KANAAL_KLANTCONTACT = Kanaal(
+    "klantcontacten",
+    main_resource=Klantcontact,
+    kenmerken=(
+        "nummer",
```
stevenbal: `nummer` has been marked as deprecated, so let's not add that here, also those kind of identifiers vary per klantcontact, so that's probably not really something applications want to subscribe to

```suggestion
```

ruleset.yaml
```diff
@@ -1,5 +1,8 @@
 # 2.1 ruleset got changed without a version increase
 extends: https://static.developer.overheid.nl/adr/2.1/ruleset.yaml
 rules:
-  nlgov:query-keys-camel-case: warn
-  nlgov:paths-kebab-case: warn
+  # Workaround for https://github.com/open-zaak/open-zaak/issues/2390
+  # until a new version of spectral-rulesets is released
+  # See also: https://github.com/stoplightio/spectral/issues/2959#issuecomment-4526418689
+  duplicated-entry-in-enum: off
```
stevenbal: You mentioned this might not be needed anymore, due to a new spectral-rulesets release, right?

```suggestion
```

src/openklant/components/klantinteracties/models/klantcontacten.py
```diff
@@ -82,6 +82,21 @@ class Klantcontact(APIMixin, models.Model):
         ),
         null=True,
     )
+    hoofd_onderwerp_type = models.URLField(
+        _("hoofd onderwerp type"),
+        help_text=_("Het type van het onderwerp van het contact"),
```
stevenbal:
```suggestion
        help_text=_("Het type van het onderwerp van het contact (bijvoorbeeld een zaaktype)"),
```

src/openklant/components/klantinteracties/api/tests/test_notifications.py
```diff
@@ -285,3 +285,129 @@ def test_send_notification_delete_object(self, m):
...
+                "kanaal": "klantcontacten",
+                "hoofdObject": f"http://testserver{self.detail_url}",
+                "resource": "klantcontact",
+                "resourceUrl": f"http://testserver{self.detail_url}",
+                "actie": "destroy",
+                "aanmaakdatum": "2024-02-02T00:00:00Z",
+                "kenmerken": {
+                    "nummer": "123456789",
+                    "hoofdOnderwerpType": None,
+                    "indicatieContactGelukt": "True",
```
stevenbal: I'm not sure if we've had other boolean kenmerken before, but could you make sure they are properly serialized to `true` or `false` instead of string values?

stevenbal (commented):
@Floris272 could you also make sure the notification docs are updated? You can use the command from notifications-api-common `src/manage.py generate_notificaties --output-file src/notificaties.md` to do this.

Also the original ticket does mention the kanaal name `klanten` and has `hoofdObject: {klantenUrl}` (probably referring to partijen), but I don't think we can determine a "main" partij linked to a klantcontact (theoretically there could be any number of partijen linked to a klantcontact), and at time of creation there won't be any linked to the klantcontact yet, so I think your implementation makes more sense.

I'll double check this with Jan Brekelmans though

src/openklant/components/klantinteracties/api/viewsets/klantcontacten.py
```diff
@@ -88,7 +90,9 @@
         description="Verwijder een klant contact.",
     ),
 )
-class KlantcontactViewSet(CheckQueryParamsMixin, ExpandMixin, viewsets.ModelViewSet):
+class KlantcontactViewSet(
+    CheckQueryParamsMixin, ExpandMixin, NotificationViewSetMixin, viewsets.ModelViewSet
```
stevenbal, line 95: In Open Zaak the convenience endpoint also emits the old style notifications, right? Could you make sure `maak-klantcontact` does this too?

src/openklant/components/klantinteracties/models/klantcontacten.py
Floris272: - systeemOnderneemtActie
- actieNodig (inverted)

src/openklant/components/klantinteracties/api/tests/test_notifications.py
Floris272: https://github.com/maykinmedia/notifications-api-common/pull/77

ruleset.yaml
Floris272: No it is stil an issues since spectral itself has not been updated but OK does not have the error. (zaken spec in OZ does)

src/openklant/components/klantinteracties/models/klantcontacten.py
Floris272: Discussed with Jan and we would like to keep verder_actie_ondernomen. Inverting the name (actie nodig) would not be correct because the system that creates the klantcontact should only say something about itself. Did add a constraint for systeemOnderneemtActie being true not being allowed when contact was successful.

src/openklant/components/klantinteracties/api/viewsets/klantcontacten.py
Floris272: added

src/notificaties.md
```diff
@@ -46,9 +69,9 @@ De architectuur van de notificaties staat beschreven op <a href="https://github.
 
 **Kenmerken**
 
-* `nummer`: Uniek identificerend nummer dat tijdens communicatie tussen mensen kan worden gebruikt om de specifieke partij aan te duiden.
-* `interneNotitie`: Mededelingen, aantekeningen of bijzonderheden over de partij, bedoeld voor intern gebruik.
-* `soortPartij`: Geeft aan van welke specifieke soort partij sprake is.
+* `nummer`: **DEPRECATED** Uniek identificerend nummer dat tijdens communicatie tussen mensen kan worden gebruikt om de specifieke partij aan te duiden.
+* `interne_notitie`: Mededelingen, aantekeningen of bijzonderheden over de partij, bedoeld voor intern gebruik.
```
stevenbal: The camelCasing seems to be removed here, not sure why, but probably related to changes in the library? The actual kenmerken are camelCased though, so this should be fixed

src/openklant/conf/base.py
```diff
@@ -53,6 +53,22 @@
     ),
 )
 
+LOG_NOTIFICATIONS_IN_DB = config(
```
stevenbal, line 57: Could you also mention https://github.com/maykinmedia/open-object/issues/757 in the commit that updates notifications-api-common? Otherwise we will forget to add this in the changelog

src/openklant/utils/notifications.py
```diff
@@ -0,0 +1,66 @@
...
+
+from django.conf import settings
+from django.db import models, transaction
+
+from notifications_api_common.models import FailedNotification, NotificationTypes
+from notifications_api_common.tasks import send_notification
+from notifications_api_common.viewsets import NotificationCreateMixin, NotificationMixin
+
+
+def create_failed_notification(message: dict) -> int | None:
```
stevenbal: Is this not something that should live in notifications-api-common?

src/openklant/components/klantinteracties/models/klantcontacten.py
```diff
@@ -82,6 +83,22 @@ class Klantcontact(APIMixin, models.Model):
...
+        ),
+        blank=True,
+    )
+
+    verdere_actie_ondernomen = models.BooleanField(
+        _("verdere actie ondernomen"),
+        help_text=_(
+            "Geeft aan of het systeem waar het contact heeft plaats gevonden zelf actie onderneemt als het niet lukt om de klant te bereiken."
+        ),
+        default=False,
```
stevenbal, line 99: For some reason the default for this doesn't seem to show up in the API spec:

Image

src/openklant/components/klantinteracties/api/tests/test_klantcontacten.py
```diff
@@ -161,6 +161,23 @@ def test_create_klantcontact(self):
...
+            "inhoud": "inhoud",
+            "indicatieContactGelukt": True,
+            "verdereActieOndernomen": True,
+            "taal": "ndl",
+            "vertrouwelijk": True,
+            "plaatsgevondenOp": "2019-08-24T14:15:22Z",
+        }
+        response = self.client.post(list_url, data)
+
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
```
stevenbal, line 179: Could you add an assertion on the error code as well?

src/openklant/components/klantinteracties/models/klantcontacten.py
```diff
@@ -82,6 +83,22 @@ class Klantcontact(APIMixin, models.Model):
...
+        help_text=_(
+            "Het type van het onderwerp van het contact (bijvoorbeeld een zaaktype)"
+        ),
+        blank=True,
+    )
+
+    verdere_actie_ondernomen = models.BooleanField(
+        _("verdere actie ondernomen"),
+        help_text=_(
+            "Geeft aan of het systeem waar het contact heeft plaats gevonden zelf actie onderneemt als het niet lukt om de klant te bereiken."
```
stevenbal, line 97: Is this helptext still correct after discussing with Jan?

stevenbal (changes requested):
Some small things remaining

src/notificaties.md
Floris272: Seems to never have been camelcase, added it in https://github.com/maykinmedia/notifications-api-common/pull/82

src/openklant/components/klantinteracties/models/klantcontacten.py
Floris272: I dont think rest framework ever did this. Should i just add it to the help text?

src/openklant/components/klantinteracties/models/klantcontacten.py
Floris272: yes? from the issue:
> Used in case of a failed klantcontact. Defines if the system which created the klantcontact will undertake further remediation

src/openklant/components/klantinteracties/models/klantcontacten.py
stevenbal: Hmm you're right, drf-spectacular doesn't seem to automatically infer the default from the model fields. I added a default via `extra_kwargs` instead.

stevenbal (approved):

stevenbal:
@Floris272 I pushed some small changes to add the explicit default for `verdereActieOndernomen`, regenerated the schema and changed the validation error to use dashes instead of underscores (to be consistent with other error codes)

PR #617, opened 2026-06-30
Feature/359 add klantcontact filters

### LINKED ISSUE #359: Additional query parameters for efficient searching of klantcontacten by digitaleAdressen
### Thema / Theme

Klantinteracties API

**Context**
In KISS / PodiumD Contact there is a use case where the employee can search for a specific `klantcontact` based on the `digitaleAdressen` that have been provided by the `betrokkene`. Specifically, only klantcontacten matching the following conditions are shown:

- The `betrokkene` specified `digitaleAdressen` matching the search query by the employee
- The `klantcontact` resulted in `interneTaken`
- The `betrokkene` was NOT a `partij`

These klantcontacten need to be sorted descending on `plaatsgevondenOp` (I believe this is the default behavior for the api endpoint for klantcontacten).

Right now, the most efficient way to do this seems to be to

- Start at `/digitaleadressen`
- Use `adres__icontains` to filter on the search query by the employee
- Expand to `verstrektDoorBetrokkene.hadKlantcontact.leiddeTotInterneTaken`
- Aggregate all pages and deduplicate all klantcontacten (client-side)
- Filter out klantcontacten without interneTaken (client-side)
- Filter out klantcontacten with a betrokkenen that is a partij (client-side)
- Sort on `plaatsgevondenOp` (client-side)

**Possible solution**
The following query parameters on the `/klantcontacten` endpoint would be a big improvement for this use case. In order of priority:
1. `hadBetrokkene__verstrektedigitaalAdres__adres__icontains`: this would allow us to use the pagination and sorting of the api endpoint as-is, removing the need to aggregate all pages and deduplicate klantcontacten. We would still need to filter out specific records, possibly resulting in inconsistent page sizes. This can be solved by the next point:
2. `hadBetrokkene__wasPartij=false` and `leiddeTotInterneTaken=true`: this would remove the need to do any filtering client-side, allowing us to offer consistent page sizes.

KCC employees handling questions about customers regarding existing klantcontacten would be able to find the klantcontact they are looking for a lot quicker.

_No response_

TimdeBeer1:
Fixes #359 

**Changes**

Added filters to klantcontact

**Checklist**

- Commit hygiene

  - [x] Commit messages refer to the relevant Github issue
  - [x] Commit messages explain the "why" of change, not the how

- Architectural design record

  - [x] If a design decision was made that changes functionality, create an architectural design record for it [here](https://github.com/maykinmedia/open-klant/wiki/Architectural-Decision-Records-(ADR))

### DIFF
--- a/src/openklant/components/klantinteracties/api/filterset/klantcontacten.py
+++ b/src/openklant/components/klantinteracties/api/filterset/klantcontacten.py
@@ -154,40 +154,36 @@ class KlantcontactFilterSet(FilterSet):
     def filter_betrokkene_partij_identificator_code_objecttype(
         self, queryset, name, value
     ):
-        try:
-            return queryset.filter(
-                betrokkene__partij__partijidentificator__partij_identificator_code_objecttype=value
-            )
-        except ValueError:
-            return queryset.none()
+        subquery = Betrokkene.objects.filter(
+            klantcontact_id=OuterRef("pk"),
+            partij__partijidentificator__partij_identificator_code_objecttype=value,
+        )
+        return queryset.filter(Exists(subquery))
 
     def filter_betrokkene_partij_identificator_code_soort_object_id(
         self, queryset, name, value
     ):
-        try:
-            return queryset.filter(
-                betrokkene__partij__partijidentificator__partij_identificator_code_soort_object_id=value
-            )
-        except ValueError:
-            return queryset.none()
+        subquery = Betrokkene.objects.filter(
+            klantcontact_id=OuterRef("pk"),
+            partij__partijidentificator__partij_identificator_code_soort_object_id=value,
+        )
+        return queryset.filter(Exists(subquery))
 
     def filter_betrokkene_partij_identificator_object_id(self, queryset, name, value):
-        try:
-            return queryset.filter(
-                betrokkene__partij__partijidentificator__partij_identificator_object_id=value
-            )
-        except ValueError:
-            return queryset.none()
+        subquery = Betrokkene.objects.filter(
+            klantcontact_id=OuterRef("pk"),
+            partij__partijidentificator__partij_identificator_object_id=value,
+        )
+        return queryset.filter(Exists(subquery))
 
     def filter_betrokkene_partij_identificator_code_register(
         self, queryset, name, value
     ):
-        try:
-            return queryset.filter(
-                betrokkene__partij__partijidentificator__partij_identificator_code_register=value
-            )
-        except ValueError:
-            return queryset.none()
+        subquery = Betrokkene.objects.filter(
+            klantcontact_id=OuterRef("pk"),
+            partij__partijidentificator__partij_identificator_code_register=value,
+        )
+        return queryset.filter(Exists(subquery))
 
     def filter_betrokkene_digitaal_adres(self, queryset, name, value):
         subquery = Betrokkene.objects.filter(
--- a/src/openklant/components/klantinteracties/api/filterset/klantcontacten.py
+++ b/src/openklant/components/klantinteracties/api/filterset/klantcontacten.py
@@ -69,7 +69,7 @@ class KlantcontactFilterSet(FilterSet):
     )
     had_betrokkene__digitaaladres__adres__icontains = filters.CharFilter(
         help_text=_(
-            "Zoek klantcontact object op basis van het digitale adres van de betrokkene."
+            "Zoek klantcontact object op basis van (een deel van) het digitale adres van de betrokkene (niet hoofdlettergevoelig)."
         ),
         method="filter_betrokkene_digitaal_adres",
     )
--- a/src/openklant/components/klantinteracties/api/filterset/klantcontacten.py
+++ b/src/openklant/components/klantinteracties/api/filterset/klantcontacten.py
@@ -150,7 +150,6 @@ class KlantcontactFilterSet(FilterSet):
             return queryset.filter(Exists(subquery))
         if value is False:
             return queryset.exclude(Exists(subquery))
-        return queryset
 
     def filter_betrokkene_partij_identificator_code_objecttype(
         self, queryset, name, value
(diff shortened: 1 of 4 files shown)


src/openklant/components/klantinteracties/api/filterset/klantcontacten.py
```diff
@@ -120,6 +141,17 @@ class Meta:
             "plaatsgevonden_op",
         )
 
+    def filter_had_betrokkene_was_partij(self, queryset, name, value):
+        subquery = Betrokkene.objects.filter(
```
stevenbal, line 145: I think this filter (and probably the other two as well) can return duplicate results, e.g. if a Klantcontact has two betrokkenen that do not have wasPartij set, the same klantcontact will be returned twice.

Could you make sure this is fixed and all three of these filters are tested to not return duplicates?

stevenbal (changes requested):
Could you fix the missing coverage? https://app.codecov.io/gh/maykinmedia/open-klant/pull/617?dropdown=coverage&src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=maykinmedia

src/openklant/components/klantinteracties/api/filterset/klantcontacten.py
TimdeBeer1: I fixed one filter that gave back duplicates, with the Exist() it doesn't give the duplicates. I also tested filter_betrokkene_partij_identificator_code_register and that gives back duplicate results. Should I fix this in this pr?

stevenbal: yes, could you do it in this PR?

stevenbal (changes requested):

src/openklant/components/klantinteracties/api/filterset/klantcontacten.py
```diff
@@ -61,6 +67,18 @@ class KlantcontactFilterSet(FilterSet):
         ),
         method="filter_betrokkene_partij_identificator_code_register",
     )
+    had_betrokkene__digitaaladres__adres__icontains = filters.CharFilter(
+        help_text=_(
+            "Zoek klantcontact object op basis van het digitale adres van de betrokkene."
```
stevenbal:
```suggestion
            "Zoek klantcontact object op basis van (een deel van) het digitale adres van de betrokkene (niet hoofdlettergevoelig)."
```

src/openklant/components/klantinteracties/api/filterset/klantcontacten.py
```diff
@@ -120,43 +141,64 @@ class Meta:
...
     def filter_betrokkene_partij_identificator_code_register(
         self, queryset, name, value
     ):
-        try:
-            return queryset.filter(
-                betrokkene__partij__partijidentificator__partij_identificator_code_register=value
-            )
-        except ValueError:
-            return queryset.none()
+        subquery = Betrokkene.objects.filter(
```
stevenbal, line 182: Could you explicitly mention in one of the commits that the existing filters previously gave duplicates and that that was fixed? Otherwise we can't see this when creating the changelog for the next release

src/openklant/components/klantinteracties/api/tests/test_filters.py
```diff
@@ -137,6 +137,41 @@ def test_filter_partij_partij_identificator_code_objecttype(self):
...
+        )
+        BetrokkeneFactory.create(
+            klantcontact=klantcontact,
+            partij=partij,
+        )
+
+        response = self.client.get(
+            self.url,
+            {
+                "hadBetrokkene__wasPartij__partijIdentificator__codeObjecttype": "natuurlijk_persoon",
```
stevenbal: Applies to the other tests as well, could you use the values from the corresponding choices classes? E.g. `PartijIdentificatorCodeObjectType.natuurlijk_persoon`

src/openklant/components/klantinteracties/api/tests/test_filters.py
```diff
@@ -256,6 +359,228 @@ def test_filter_partij_partij_identificator_code_register(self):
...
+            },
+        )
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+        data = response.json()["results"]
+
+        self.assertEqual(1, len(data))
+        self.assertEqual(str(klantcontact.uuid), data[0]["uuid"])
+
+    def test_filter_verstrektedigitaal_adres(self):
```
stevenbal:
```suggestion
    def test_filter_hadBetrokkene__digitaaladres__adres__icontains(self):
```

src/openklant/components/klantinteracties/api/tests/test_filters.py
```diff
@@ -256,6 +359,228 @@ def test_filter_partij_partij_identificator_code_register(self):
...
+        with self.subTest("no_matches_found_return_nothing"):
+            response = self.client.get(
+                self.url,
+                {"hadBetrokkene__digitaaladres__adres__icontains": "does-not-exist"},
+            )
+
+            self.assertEqual(response.status_code, status.HTTP_200_OK)
+            self.assertEqual(response.json()["count"], 0)
+
+    def test_filter_verstrektedigitaal_adres_no_duplicates(self):
```
stevenbal:
```suggestion
    def test_filter_hadBetrokkene__digitaaladres__adres__icontains_no_duplicates(self):
```

stevenbal (changes requested):
Code looks good, some small comments remaining

stevenbal (approved):
@TimdeBeer1 these were the existing filters that could return duplicates that were fixed right?
- hadBetrokkene__wasPartij__partijIdentificator__codeObjecttype
- hadBetrokkene__wasPartij__partijIdentificator__codeRegister
- hadBetrokkene__wasPartij__partijIdentificator__codeSoortObjectId
- hadBetrokkene__wasPartij__partijIdentificator__objectId

I think there are other filtersets that implement the same filters, which are affected by that issue as well (like DigitaleAdressen). Could you make a separate issue to fix those?

TimdeBeer1:
> @TimdeBeer1 these were the existing filters that could return duplicates that were fixed right?
> 
>     * hadBetrokkene__wasPartij__partijIdentificator__codeObjecttype
> 
>     * hadBetrokkene__wasPartij__partijIdentificator__codeRegister
> 
>     * hadBetrokkene__wasPartij__partijIdentificator__codeSoortObjectId
> 
>     * hadBetrokkene__wasPartij__partijIdentificator__objectId
> 
> 
> I think there are other filtersets that implement the same filters, which are affected by that issue as well (like DigitaleAdressen). Could you make a separate issue to fix those?

Yes, right. I created a new issue to fix those: https://github.com/maykinmedia/open-klant/issues/624

PR #629, opened 2026-07-14
:arrow_up: Bump maykin-common==0.20.1

danielmursa-dev:
Fixes Taiga Issue https://taiga.maykinmedia.nl/project/maykin-intranet/issue/1516

Closes https://github.com/maykinmedia/open-api-framework/issues/223 partially

**Changes**

[Describe the changes here]

**Checklist**

- Commit hygiene

  - [ ] Commit messages refer to the relevant Github issue
  - [ ] Commit messages explain the "why" of change, not the how

- Architectural design record

  - [ ] If a design decision was made that changes functionality, create an architectural design record for it [here](https://github.com/maykinmedia/open-klant/wiki/Architectural-Decision-Records-(ADR))


src/openklant/conf/base.py
```diff
@@ -172,3 +173,59 @@
...
+            "CUSTOM_PRODUCT_URL",
+            default="",
+            documentation=DocumentationParams(
+                help_text=(
+                    "Optional link for the custom product when redistributing the "
+                    "application. If provided, the product name will be clickable."
+                ),
+                group="Branding",
+            ),
+        ),
```
stevenbal: The params defined inside this block are currently not picked up in the envvar docs, because at runtime when generating docs they are not called (there is a conditional on `custom_product_name`)

Image

stevenbal (changes requested):

src/openklant/conf/base.py
danielmursa-dev: Yes, thanks!

stevenbal (commented):
@danielmursa-dev do we need to make any additional changes to the dashboard page footer? Perhaps that should be part of maykin common as well? I'll check with Sergei in a bit to see what the use case for this branding is.

Image

stevenbal (approved):

